PR #8, opened 2015-01-08
dahdi-linux: fix oct612x compile on some ARM targets

dangowrt:
Unbreak dahdi-linux build at least for cns3xxx, kirkwood, mvebu and oxnas
by explicitely including slab.h in oct612x-user.c.


jslachta:
Thank you.

PR #17, opened 2015-03-08
asterisk-11.x: fix mysql_config location

plntyk:
on Arch Linux configure find host mysql_config first
set location explicitly to fix build fail


jslachta:
Thank you.

PR #18, opened 2015-03-08
asterisk-13.x: fix xml2-config location

plntyk:
on Arch Linux libxml2 detection fails
set xml2-location explicitly to fix build fail


jslachta:
Thank you.

PR #20, opened 2015-03-16
Update sccp.conf

dkgroot:
Fixing binaddr
Removing unnecessary bindaddr
Removing protocolversion
Adding a couple more sample entries in device and line


jslachta:
Thank you, Diederik. :-)

dkgroot:
No problem ;-) Would you like me to create an chan-sccp-b sqlite3 example (binary or .sql ?) as well. Or should that be left to the user ? Would also require extconfig.conf / res_config_sqlite3.conf changes.

jslachta:
That would be nice if you could do it so..

jslachta:
The extconfig.conf is a part of default asterisk package - if the changes would go to this package, it would be nice, if the changes would go apart from the asterisk base package, that would interfere with the files provided by the package.

dkgroot:
Hi Jiri,
Can't we do something like:

        if [ -z "`grep 'sccpdevice' /etc/asterisk/extconfig.conf` ]; then
                cat files/extconfig.sccp.conf >> /etc/asterisk/extconfig.conf
        fi

In the makefile ? That would do the trick nicely i would think, without specifying anything is the chan-sccp-b package is not installed. During the same check we could also find out where the use stored the sqlite3.db and update it appropriately, if none of the tables can be found. We could even ask if the database should be created during plg installation (not sure if we can ask the user anything).

PR #21, opened 2015-03-16
Update Asterisk11 Makefile (Untested)

dkgroot:
Shorten Package Description for easier readability when displaying "pkg list | grep asterisk11"


dkgroot:
Just noticed you already did something similar in f9c3f701853b8b980283177b415487e087737ed2 So i guess it works.

jslachta:
Hello Diederik,

your changes should work without a problem. Would you be so kind to do that also for asterisk18 and asterisk13? 

Thank you

PR #24, opened 2015-03-31
bcg729: source tarball MD5 changed mysteriously

dangowrt:
Anyone still got the old tarball in their download folder? PLEASE make a diff and post it here, so we can see what actually changed.


jslachta:
Hi Daniel,

would you be so kind to take a look at the diff? https://gist.github.com/jslachta/539ea1c349609776dd02

jow-:
Seems like someone just regenerated the autoconf/automake files. Random m4 crap exchanged with other random m4 crap - nothing interesting as far as I can see.

PR #40, opened 2015-04-22
Asterisk13.x: Add missing Confbridge

redfox1977:
Add the missing Confbridge and it brings the bridge_simple.so to activate the sip module.
Without the bridge_simple.so can not make calls via sip it.


jslachta:
Would you be so kind to put your real name into Signed-off-by line? Otherwise your PR seems to be OK.

redfox1977:
So OK?

PR #49, opened 2015-04-24
Asterisk 13.x: Add depensies for chan_sip

redfox1977:
Add the depensies for the chan_sip modul bacause chan_sip dos not function with out it.


jslachta:
Why chan_sip depends on app_confbridge? Is there any specific use case for it?

redfox1977:
The chan_sip needs the bridge_simple. So and these are in the app_confbridge packet. Without these so file you can not use the sip protokoll to phone with...


Von Samsung Mobile gesendet

-------- Ursprüngliche Nachricht --------
Von: Jiri Slachta <notifications@github.com> 
Datum: 28.04.2015  10:15  (GMT+01:00) 
An: openwrt/telephony <telephony@noreply.github.com> 
Cc: Sven Pietack <redfox7719@googlemail.com> 
Betreff: Re: [telephony] Asterisk 13.x: Add depensies for chan_sip (#49) 
 
Why chan_sip depends on app_confbridge? Is there any specific use case for it?

jslachta:
ok. I will test the changes and merge it, if it works. ;-)

PR #50, opened 2015-04-24
Asterisk 13.x: Sorting the modules

redfox1977:
Sort the modules in alphabetic order. It is much easyer to find missing modules in alphabetic order.


jslachta:
Modules are reordered. There were more modules in incorrect alphabetic order. Should be fixed by now.

PR #55, opened 2015-05-14
Avoid trying to load the echo module from kmod-dahdi-echocan-oslec

VittGam:
It would result in the warning "echo is already loaded" during package installation.


jslachta:
Hi @VittGam,

please add prefix to your commit (dahdi-linux: your commit short description) and add Signed-off-by line.

VittGam:
The Signed-off-by line is already there... Anyway I'm going to delete the commit and add another one with the required prefix.

PR #69, opened 2015-06-22
asterisk-11x: Update 010-asterisk-configure-undef-res-ninit.patch

micmac1:
Lately configure is recreated from configure.ac. The current patch changes configure. So we patch it, then it gets recreated and our change is gone. So patch configure.ac instead. Without this change the build fails due to an undefined reference to res_nsearch.


dangowrt:
Thanks!

PR #70, opened 2015-06-23
asterisk-11.x: Don't bootstrap menuselect

micmac1:
The configure script generated by bootstrap.sh in the menuselect folder doesn't work properly. It doesn't check for 'sed', for instance, resulting in './configure: line 4622: -e: command not found'. The build works well without the bootstrap.


dangowrt:
Thanks!

PR #71, opened 2015-06-25
miax: fix musl build

plntyk:
- error.h does not seem to be used (musl doesnt provide one)

fixes error/handling of FILE *stderr leading to errors like:
miax.c:217:2: error: assignment of read-only variable 'stderr'
  stderr=fopen(log_output,"w");

thx to jow for pointing things out


dangowrt:
Thanks!

PR #75, opened 2015-06-25
siproxd: fix musl compatibility

jow-:
- Add missing `sys/types.h` include to provide `u_short` type under musl
 - Add missing `string.h` includes to avoid implicit function declarations

Signed-off-by: Jo-Philipp Wich <jow@openwrt.org>


dangowrt:
Thanks!

PR #87, opened 2015-09-13
yate-scripts-perl: Port script banbrutes.php to perl

robho:
Yate ships with a php script (banbrutes.php) that can help block brute force
login/password attacks. This php script is now ported to perl so that it's
easier to use with openwrt.


robho:
Since you've merged #123 (thanks!) I'll close this pull request.

PR #92, opened 2015-10-08
backport fixes and minor updates from master to for-15.10

dangowrt:
Lots of fixes and minor updates have accumulated since CC 15.05 has been released.
Import everything which does affect CC 15.05 (ie. skip the musl fixes and such which are only relevant on DD).


dangowrt:
merging this now...

PR #103, opened 2016-02-06
added app_queue, func_presencestate and cfg files for CEL and HTTP

dzavy:
- app_queue was missing on the list of optional modules
- func_presencestate was missing on the list of optional modules
- CEL doesn't work without cel.conf
- HTTP server/WebSockets doesn't work without http.conf


jslachta:
Thank you, I manually merged your changes (due to other changes in asterisk Makefiles. I also added your changes to asterisk13 package.

PR #106, opened 2016-04-05
Fix menuselect build in asterisk 11.x/13.x

micmac1:


jslachta:
Are you sure with your PR? When I apply your PR it does not seem to compile while the current sources does..

micmac1:
Hello Jiri,

Maybe it fails here because I'm compiling agains stable CC, not openwrt trunk. I'll setup trunk and try with that, then report back.

Thanks!
Sebastian

micmac1:
Hello Jiri,

With trunk it works without the patches. I'll close this PR. Sorry for the noise.

Bye,
Sebastian

PR #120, opened 2016-09-09
yate: Call pthread_key_create when the key is first accessed

### LINKED ISSUE #99: yate segfaults
Original issue at https://github.com/openwrt/packages/issues/2323

yate (5.5.0-1-1)
OpenWrt Designated Driver r48135
TP-Link TL-WDR3500

When I start yate or attempt to run it from the command line, I get a segfault immediately. I have tried with yate modules installed, and with no modules at all installed. Even when passing --help, it crashes before it gets to process command-line arguements.

Kernel log output:

```
do_page_fault(): sending SIGSEGV to yate for invalid read access from 00000000
epc = 77416ce0 in libc.so[773a0000+92000]
ra  = 772c919b in libyate.so.5.5.0[77286000+e9000]
```

gdb shows that the crash is in musl's ~~tss_set~~ tss_get function, but can't get the rest of the call stack:

```
$ gdb -q yate
Reading symbols from yate...(no debugging symbols found)...done.
(gdb) r
Starting program: /usr/bin/yate

Program received signal SIGSEGV, Segmentation fault.
0x77fc8ce0 in tss_get () from /lib/ld-musl-mips-sf.so.1
(gdb) bt
#0  0x77fc8ce0 in tss_get () from /lib/ld-musl-mips-sf.so.1
warning: GDB can't find the start of the function at 0x77e7b19a.

    GDB is unable to find the start of the function at 0x77e7b19a
and thus can't determine the size of that function's stack frame.
This means that GDB may be unable to access that stack frame, or
the frames below it.
    This problem is most likely caused by an invalid program counter or
stack pointer.
    However, if you think GDB should simply search farther back
from 0x77e7b19a for code which looks like the beginning of a
function, you can increase the range of the search using the `set
heuristic-fence-post' command.
#1  0x77e7b19b in ?? ()
(gdb) set heuristic-fence-post 100000
(gdb) bt
#0  0x77fc8ce0 in tss_get () from /lib/ld-musl-mips-sf.so.1
warning: GDB can't find the start of the function at 0x77e7b19a.
#1  0x77e7b19b in ?? ()
```

~~tss_set in musl:
http://git.musl-libc.org/cgit/musl/tree/src/thread/tss_set.c~~

~~I couldn't find any code calling tss_set in musl, yate, or uclibcxx (a dependency of yate).~~ I am trying to compile a debug build of yate now. Is anyone able to reproduce this issue or have an idea of why it might be happening?

robho:
pthread_key_create used to be called when a specific static key object was
initialized, but if another static object was initialized earlier and this
other object needed the key during its initialization yate would crash.

Fixes #99

### DIFF
--- a/net/yate/patches/120-create-thread-key-on-access.patch
+++ b/net/yate/patches/120-create-thread-key-on-access.patch
@@ -1,0 +1,49 @@
+--- a/engine/Thread.cpp
++++ b/engine/Thread.cpp
+@@ -106,21 +106,18 @@ static DWORD getTls()
+     return tls_index;
+ }
+ #else /* _WINDOWS */
+-static pthread_key_t current_key;
+-
+-class ThreadPrivateKeyAlloc
++static pthread_key_t& current_key()
+ {
+-public:
+-    ThreadPrivateKeyAlloc()
+-    {
+-	if (::pthread_key_create(&current_key,ThreadPrivate::destroyFunc)) {
++    static pthread_key_t* current_key = NULL;
++    if (!current_key) {
++	current_key = new pthread_key_t;
++	if (::pthread_key_create(current_key, ThreadPrivate::destroyFunc)) {
+ 	    abortOnBug(true);
+ 	    Debug(DebugFail,"Failed to create current thread key!");
+ 	}
+     }
+-};
+-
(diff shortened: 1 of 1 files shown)


jslachta:
Thank you for the PR.

PR #122, opened 2016-10-03
yate-scripts-perl: Make perlbase-data optional

robho:
Package yate-scripts-perl requires package perlbase-data to work
properly, but perlbase-data is only needed for function dump() which I
expect to be used only when debugging YATE modules. It's possible to
write YATE modules without using the dump() function and then it's a
waste of resources to require the user to have perlbase-data installed
since it will consume memory while the YATE module is running and disk
space for the package installation.

So, let's change Yate.pm to only load Data::Dumper (from perlbase-data)
when the dump() function is called. This way Yate.pm can be used even
if perlbase-data isn't installed, but if you want to use the dump()
function you of course need to install perlbase-data.

This is an alternative to #86 and I suggest that #86 is discarded and this change is taken instead, but I'll let you make the decision.

### DIFF
--- a/net/yate/Makefile
+++ b/net/yate/Makefile
@@ -56,7 +56,7 @@ endef
 
 define Package/$(PKG_NAME)-scripts-perl
   $(call Package/yate/Default)
-  DEPENDS += $(PKG_NAME) $(PKG_NAME)-mod-extmodule +perlbase-data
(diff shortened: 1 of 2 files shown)


robho:
Since you've merged #86 (thanks!) I'm dropping this pull request.

PR #129, opened 2017-01-04
libs/re: Add zlib dependency

diizzyy:
Compile tested on: Kirkwood, LEDE trunk

Adds missing zlib dependency

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>


diizzyy:
@jslachta

PR #135, opened 2017-03-09
Feature chltq

kochstefan:
Adds improved channel "chan_lantiq" (known from asterisk 1.8) to asterisk 11 and 13.
This enables usage of FXS ports from lantiq routers by asterisk.


dangowrt:
Great work! Thanks!

PR #137, opened 2017-03-10
Add xmpp and motif modules

zorxd:
This commit adds back xmpp and motif modules which were available in asterisk 11.


jpap:
This helped me, thanks @zorxd.

@jslachta, can we get this merged in?  It's critical for Asterisk connecting with Google Voice.

jslachta:
Hi, this should be fixed in commit [6341be0](https://github.com/openwrt/telephony/commit/6341be04f4ae5ff1e03595eda5a62b17c3bf5301). Feel free to create an issue or PR if this does not work.

PR #138, opened 2017-03-11
asterisk-13.x: update to version 13.14.0

zorxd:


jslachta:
Thank you for the PR. I tested it and it works.

PR #144, opened 2017-05-05
freeswitch-stable-sounds: General fixes

micmac1:
- Fixed sha256 checksums of some of the Russian language prompts
- Commented out non 8 kHz sound packs to not clog the mirrors
- Changed versioning scheme to the one used by upstream
- Made "Sounds" function params more generic


micmac1:
@jow-

PR #145, opened 2017-05-07
freeswitch-stable-sounds: add bigger sound packs (tagged BROKEN)

micmac1:
As suggested by Jo I uncommented the big sound packs, but marked as
BROKEN, meaning they're still available for selection, but not built by
default.


micmac1:
Hello @jow-

jslachta:
Thank you @micmac1 , this is preferable solution to use it that way.

PR #146, opened 2017-05-09
Fix PKG_SOURCE_URL for libpri 1.4.15

pho:
The 1.4.15 file has been moved into an /old/ subfolder


jslachta:
Thank you for the PR.

PR #149, opened 2017-05-24
PR for freeswitch-stable

micmac1:
Hello all,

Since the last batch of commits the package doesn't compile anymore because of a wrong host swig location. I hadn't realized it before because on my own computer the swig location was still correct. Anyway, sorry for that. Here is the fix and some other changes.

Thanks!
Sebastian


jow-:
Merged, thanks @micmac1 !

PR #154, opened 2017-06-02
PR for freeswitch-stable

micmac1:
Now that the swig location is fixed the build dies someplace else, at erlang. I (hopefully) fixed erlang, and I assume with python we would run into the same issue, so here a PR with an erlang and a python fix.

@jow- I know you're busy, but maybe if you get around to it... :) Thank you


jow-:
Merged, thanks.

PR #170, opened 2017-07-22
New PR: mostly Asterisk (bumps, fixes, new package too)

micmac1:
Hello Jiri @jslachta 

I wanted to add a module that links against bcg729, for which I needed a version bump. So I updated the asterisk module along with it. Asterisk didn't really compile in trunk so I went and fixed what I found until it compiled and I could compile-test asterisk-g72x.

I read in some of the issue reports that asterisk has some issues loading modules on musl (missing symbols), but I couldn't reproduce that. I loaded a few modules and it worked fine.

Best regards,
Sebastian

### DIFF
--- a/libs/bcg729/Makefile
+++ b/libs/bcg729/Makefile
@@ -1,5 +1,5 @@
 # 
-# Copyright (C) 2006-2014 OpenWrt.org
+# Copyright (C) 2006-2017 OpenWrt.org
 #
 # This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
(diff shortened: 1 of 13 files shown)


micmac1:
Hello Jiri @jslachta 

While checking into the chan-sccp-b missing symbols issue I saw that the ASTCFLAGS/ASTLDFLAGS weren't picked up by "make" in both asterisk-11.x and asterisk-13.x. I added a commit to this PR to fix this.

Kind regards,
Sebastian

jslachta:
Thank you so much for the fixes, Sebastian @micmac1.

PR #172, opened 2017-07-28
New PR: asterisk

micmac1:
Hello Jiri @jslachta 

It took a while to come up with this because the buildbots didn't build much for a few days. But now they're running again and I looked around what still needs fixing.

I think this should do it.

Thanks!
Sebastian


micmac1:
Hello Jiri @jslachta 

I've added another commit to add two security fixes provided by Asterisk project to pjproject.

Best regards,
Sebastian

jslachta:
Thank you, @micmac1, for your good work.

PR #180, opened 2017-08-14
chan-sccp-b: switch to git version, update patch

micmac1:
- As suggested by upstream in issue #167, move from svn to git source.
  The resulting module is less than half the size in comparison.

- Rework (and rename) previous "drop-Og-optimization" patch. The
  previous patch worked fine with -Os in CFLAGS (default). But if numeric
  values are used instead (e.g. -O2) this optimization would be completely
  removed. The new patch prevents that.


jslachta:
Thank you, @micmac1. I have compile and run tested it, everything looks ok.

micmac1:
Thank you Jiri!

Am 15. August 2017 17:28:38 MESZ schrieb Jiri Slachta <notifications@github.com>:
>Thank you, @micmac1. I have compile and run tested it, everything looks
>ok.

Issue #185, opened 2017-09-10
asterisk11/13 AMI not working


Infactum:
LEDE  Reboot (17.01.2, r3435-65eec8bd5f)
NetGear WNDR3800 (also tested on x86 build)

Typical asterisk11 (or 13) installation. The only config change is `enabled=yes` in `manager.conf`.

```
Asterisk 13.9.1, Copyright (C) 1999 - 2014, Digium, Inc. and others.                           
Created by Mark Spencer <markster@digium.com>                                                  
Asterisk comes with ABSOLUTELY NO WARRANTY; type 'core show warranty' for details.             
This is free software, with components licensed under the GNU General Public                   
License version 2 and other licenses; you are welcome to redistribute it under                 
certain conditions. Type 'core show license' for details.                                      
=========================================================================                      
Connected to Asterisk 13.9.1 currently running on LEDE (pid = 10244)                           
No entry for terminal type "cygwin";                                                           
using dumb terminal settings.                                                                  
LEDE*CLI> !nc localhost 5038                                                                   
[Sep 10 12:13:25] WARNING[10315]: tcptls.c:684 handle_tcptls_connection: FILE * open failed!   
                                                                                               
LEDE*CLI>                                                                                      
```

No TLS features are enabled.

Infactum:
Found more info in other closed issue:
https://bugs.alpinelinux.org/issues/6394
So is there any chance to hotfix this problem in asterisk11/13 ?

jslachta:
This hotfix should be fixed, please try it and consider closing this issue if resolved.

Infactum:
Can't test on router right now, but it works OK on x86.

s2504s:
I get the same issue on 
```
OWRT_TAG_NAME="v17.01.4"
```
and 
`config/feeds.conf.default`

```
src-git telephony https://git.lede-project.org/feed/telephony.git;lede-17.01
```

micmac1:
Hello Sergey,

If I remember correctly that was iostreams related (or the lack thereof). Initially there was a backport patch for 13. But it was too big and prevented updates, so it was removed.

I think 15 has proper iostreams support. So if you move to openwrt 18.06 and asterisk 15 it'll probably work.

Regards
Seb

PR #186, opened 2017-09-10
New PR: asterisk 11 + 13 (cleanups, fixes and version bumps)

micmac1:
Hi Jiri @jslachta 

This basically does the same cleanups to 11 that were done to 13. Couldn't be done 100% the same because in 11 the bundled pjproject would get in the way a bit otherwise.

For 11 it also turned out that libedit needs to stay internal :( Anyway...

For 13 I added a patch from alpine linux to fix issue #185.

Regarding the recently spotted segfault in 11 (in res_rtp_asterisk.so) I don't know. I started compiling with debug enabled but then the segfault went away.

Kind regards,
Seb

### DIFF
--- a/net/asterisk-11.x/Makefile
+++ b/net/asterisk-11.x/Makefile
@@ -63,7 +63,7 @@ define BuildAsterisk11Module
   endef
 
   define Package/asterisk11-$(1)/conffiles
-$(5)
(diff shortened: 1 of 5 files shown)


jslachta:
ACK, thank you for the fixes and the cleanup.

PR #187, opened 2017-09-14
New PR: asterisk chan-dongle and chan-lantiq

micmac1:
Hello Jiri,

This is the PR I mentioned yesterday. It wasn't time-consuming after all. chan-lantiq builds locally in nonshared mode, so it seems to work. I find the package here:

ls -lh bin/targets/lantiq/xrx200/packages/asterisk13-chan-lantiq_1.0-1_mips_24kc.ipk 
-rw-r--r-- 1 sk users 20K 14. Sep 20:33 bin/targets/lantiq/xrx200/packages/asterisk13-chan-lantiq_1.0-1_mips_24kc.ipk

Best regards,
Seb


jslachta:
It seems to be working, we will see whether the buildbots are going to build the chan-lantiq package. Thank you.

PR #188, opened 2017-09-27
New PR: asterisk version bumps plus chan-lantiq changes

micmac1:
Hello Jiri @jslachta 

The version bumps are straightforward, but the chan-lantiq relates commits (3 and 4) are a bit "out there".

I thought about how to solve the hen and egg problem for the chan-lantiq nonshared builds. I had 3 ideas, number 1 and 2 were completely ugly.

Idea 3 is in the commits 3 and 4. Basically the asterisk package puts the asterisk headers into an archive and stores it in $(DL_DIR). Then chan-lantiq gets the headers from there and builds the module.

What I don't like about this is obviously that the asterisk package misuses the download directory as a storage area. What I also don't like is that I had to rename the chan-lantiq module package a bit as otherwise it would mess with menuconfig.

On the other hand, while the above isn't particularly beautiful, it isn't anything sinister either. And the goal of users not having to build chan-lantiq manually would be accomplished. I think the only way to solve this issue with the build system we have is to resort to some sort of hackishness like this.

If you decide against this, no problem, I get it. Maybe just cherry-pick the version bumps then.

Best regards,
Seb


jslachta:
Hi @micmac1,

I am sorry for the delayed response, I am thinking about it for several days and have not found solution yet. Thank you for your proposal. You have proposed quite a good solution, but I do not like the idea that the package depends on the generated asterisk headers archive.

Let me think about it one more time. If I do not come up with a solution, I will pick yours.

micmac1:
Hello Jiri @jslachta 

I added some more commits. Same sentiment as before: if you don't want the lantiq changes maybe you can cherry-pick some of these.

Thanks!
Seb

jslachta:
I am merging this. We will come up with a better solution later.

micmac1:
Cool, I hope this actually works in the meantime. Thanks!

Am 8. Oktober 2017 16:41:56 MESZ schrieb Jiri Slachta <notifications@github.com>:
>I am merging this. We will come up with a better solution later.

PR #196, opened 2017-11-11
CC: re/rem: fix build on CC and upgrade

wigyori:
Maintainer: @jslachta 
Compile-tested on: CC/adm8668

Description:
As the earlier versions were pulled from the download site, bump re/rem packages to more recent versions.


jslachta:
Thank you, @wigyori

Issue #193, opened 2017-10-20
Chan Lantiq still not build.


Bongo79:
I read this.
http://lists.infradead.org/pipermail/lede-dev/2017-September/008971.html
http://lists.infradead.org/pipermail/lede-dev/2017-October/009279.html

what is happen when at the time where building chan-lantiq the asterisk-13.x package where not ready build yet ?
The complete depends are this: @(TARGET_lantiq_falcon||TARGET_lantiq_xway||TARGET_lantiq_xrx200) +kmod-ltq-vmmc +kmod-ltq-tapi +asterisk13
right ?

If i you I will try the ready build $(CONFIG_TARGET_ARCH_PACKAGES))-headers-for-chan-lantiq-$(VERSION_NUMBER).tar.xz inside your package.
Of course it is not the fine english Art but better as nothing.

micmac1:
Hi @Bongo79 

Depends in trunk are: DEPENDS:=@(TARGET_lantiq_falcon||TARGET_lantiq_xway||TARGET_lantiq_xrx200) +kmod-ltq-vmmc. The dep on asterisk13 was removed.

And yes, $(CONFIG_TARGET_ARCH_PACKAGES))-headers-for-chan-lantiq-$(VERSION_NUMBER).tar.xz is a hack. When you check https://sources.lede-project.org/ you see that the file asterisk13-mips_24kc-headers-for-chan-lantiq-SNAPSHOT.tar.xz actually exists. But still the hack doesn't work.

Sorry, I don't know why it is failing. To tell you the truth I'm tempted to send a pull request to get all the changes I've sent for chan-lantiq reverted. There is simply no benefit as the package still isn't built. And the way it was originally, at least the package had a sane name and didn't depend on asterisk header archive to be available.

If nobody else has an idea then I'll send that pull request.

Bye,
Sebastian

Bongo79:
OK but try to add **+kmod-ltq-tapi** first
because inside: https://github.com/kochstefan/asterisk_channel_lantiq/blob/master/src/channels/chan_lantiq.c

> /* Lantiq TAPI includes */
> #include <drv_tapi/drv_tapi_io.h>
> #include <drv_vmmc/vmmc_io.h>

micmac1:
Hi there,

I doubt adding a dep on kmod-ltq-tapi would have any benefit, because the package depends on  kmod-ltq-vmmc, which in turn depends on kmod-ltq-tapi.

Kind regards,
Sebastian

[...]

PR #199, opened 2017-11-12
asterisk-chan-lantiq: remove nonshared hack

micmac1:
Hello @jslachta 

I'm not sure what you got queued up for chan-lantiq. If this doesn't get in the way of anything please consider merging it.

Kind regards,
Seb

The previously conceived hack - to make the buildbots actually build
this package during the target phase - does not work. This patch removes
the hack. This includes removing the lines from the asterisk Makefiles
that tar up the asterisk headers.

The upside of removing the hack is that the package name can be reverted
to what it was before, so it can be found in the same place like all the
other asterisk modules, i.e. below its asterisk version.

Another upside is that the Makefile does not pretend to solve the
problem anymore. Issue #193 is still open for the same. Maybe some day
somebody can figure out how to resolve the problem properly.


jslachta:
Thank you, @micmac1.

PR #201, opened 2017-11-14
New PR: libsrtp2, pjproject and asterisk-13/15

micmac1:
Hello Jiri @jslachta 

I updated pjproject and while at it checked if libsrtp could be replaced by libsrtp2 there and in asterisk as well. It all compiled fine with the libsrtp2 update you pushed a short while ago. There are two packages I found that still depend on libsrtp: freeswitch and gst1-plugins-bad.

Best regards,
Seb


jslachta:
Thank you, Sebastian @micmac1. ACK

PR #206, opened 2017-11-23
chan-dongle: add variant for asterisk 15

micmac1:
chan-dongle already supports asterisk 15. This commit adds a variant for
it.


jslachta:
Thank you!

PR #209, opened 2017-11-23
chan-lantiq: setup Lantiq kernel headers locally

micmac1:
Hi Jiri,

This commit unpacks the sources of lib_ifxos, drv_tapi and drv_vmmc,
patches them and sets up their header files for the compiler to use when
building chan-lantiq.

The dependency on kmod-ltq-vmmc is dropped so that the package can be
built by the buildbots. Instead, a dependency on mips is added so that
the buildbots will only build the package for the ARCH that Lantiq
actually runs on.

### DIFF
--- a/net/asterisk-chan-lantiq/Makefile
+++ b/net/asterisk-chan-lantiq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=chan-lantiq
 PKG_VERSION:=20171111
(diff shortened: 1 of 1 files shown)


jslachta:
Hi Sebastian @micmac1,

I like how you approached to the building issue. I understand how difficult is to build this on buildbot, even if I do not like the idea of having a package/module without strong dependency to the asterisk that uses it.

Thank you for your hard work.

Kind regards,
Jiri

PR #213, opened 2017-11-26
New PR: re, rem, restund and baresip

micmac1:
Hello Jiri @jslachta 

These commits change the Build/Compile routine to the default one for all four packages. This has the benefit of using the standard features like FORTIFY_SOURCE and RELRO. Also cleanups where done, modules added and the build failure of baresip that is currently happening is also fixed.

Best regards,
Seb


jslachta:
Thank you, Sebastian. I already had those fixes done locally, but you fixed and enhanced much more things, thus I prefer your changes. +1 perfect work

PR #215, opened 2017-11-28
Asterisk: Prevent building of all default modules

micmac1:
Hi Jiri.

I think this is a cool feature. First it disables all modules that get enabled by default, then it enables the modules required for the selected packages. Benefits are that stuff we don't need for the packages isn't built (so the build takes less time) and that there are modules that might be added in the future that are not enabled by default and this mechanism automatically takes care of enabling them.

This doesn't touch the compiler settings or sound files. Just the items that are actually compiled.

Kind regards,
Seb


jslachta:
Ack from me. Thank you.

PR #220, opened 2017-12-05
restund: remove build timestamp

lynxis:
Build timestamps prevent reproducible builds [0].

[0] https://reproducible-builds.org/docs/timestamps/

Signed-off-by: Alexander Couzens <lynxis@fe80.eu>


lynxis:
@dangowrt @lperkov

PR #237, opened 2018-01-11
Asterisk part I for 15.05

micmac1:
Maintainer: @jslachta 
Compile tested: mips_24kc
Run tested: -

Description:

Hi Jiri!

I checked the open CVEs for Asterisk 11 and 13 and started gathering the patches. For Asterisk 11 the patches from upstream apply without failure, but for Asterisk 13 I was not so lucky.

I found patches for 12 CVEs, The first few applied OK, but then I got lots of hunk failures. It was not possible for me to backport the patches (I did not find another distro with backported fixes either). So I decided I would instead update Asterisk 13 to a recent version instead. I also added some fixes that are already in master on top.

Also included are post-install notes for Asterisk 1.8 and 11, warning that they're end of live and that an upgrade should be done to 13.

I understand it's nice to stick to a specific version in a release. But in this case I think it's good to update to a newer version. It'll make maintaining (adding future patches) easier and it provides a viable upgrade path for people who use Asterisk 11 (or even 1.8) - to a version with all CVEs fixed.

I also updated the external modules like chan-dongle, so nobody can complain that a module is missing that was available for Asterisk 11. I didn't want to put too much into this PR, so I'll send another for that.

If this is to your liking then I'd do the same for lede-17.01. In the meantime I'll check the rest of the 15.05 packages.

Kind regards,
Sebastian


micmac1:
Oh, forgot to mention. For pjproject it is the same situation. I found 4 CVEs, for the first two upstream patches apply OK, but for the last two they don't. Debian does not have backported patches for these either, their older versions are still listed as vulnerable on their CVE tracker:

https://security-tracker.debian.org/tracker/CVE-2017-16875
https://security-tracker.debian.org/tracker/CVE-2017-16872

Issue #35, opened 2015-04-17
freeswitch error


taeasy:
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... yes
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether /home/main/x86/staging_dir/toolchain-i386_i486_gcc-4.8-linaro_uClibc-0.9.33.2/bin/i486-openwrt-linux-uclibc-gcc accepts -g... yes
checking for /home/main/x86/staging_dir/toolchain-i386_i486_gcc-4.8-linaro_uClibc-0.9.33.2/bin/i486-openwrt-linux-uclibc-gcc option to accept ISO C89... none needed
checking how to run the C preprocessor... /home/main/x86/staging_dir/toolchain-i386_i486_gcc-4.8-linaro_uClibc-0.9.33.2/bin/i486-openwrt-linux-uclibc-gcc -E
checking for grep that handles long lines and -e... /home/main/x86/staging_dir/host/bin/grep
checking for egrep... /home/main/x86/staging_dir/host/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking for ldap support...
checking for default DBM... sdbm (default)
checking for dbd/apr_dbd_mysql.c... configure: error: cannot check for file existence when cross compiling
configure: error: ./configure.gnu failed for libs/apr-util
make[3]: **\* [/home/main/x86/build_dir/target-i386_i486_uClibc-0.9.33.2/freeswitch-1.5.14/.configured_yyyyyynnyynnnynnyynynnnyyynynyyyyyyyyyyyyynyynyyyyyyyyyyyyyyyyynyyyyyyynyynyyyynynynyyyynyyyyyyynnnnyyyyynnnynnyyyyynyyyyyyyyyyyyyyyyynyyyyyyyyynnnnnnyyyyyyynynyyyyyyyynynyyyyyyyyn] Error 1
make[3]: Leaving directory `/home/main/x86/feeds/telephony/net/freeswitch'
make[2]: *** [package/feeds/telephony/freeswitch/compile] Error 2
make[2]: Leaving directory`/home/main/x86'
[...]

Issue #10, opened 2015-01-20
freeswitch broken in buildbot


champtar:
Hi
An issue was open in package feed for freeswitch (openwrt/packages#630)
see also
http://buildbot.openwrt.org:8010/broken_packages/ar71xx/freeswitch/compile.txt

plntyk:
Reason why it's broken: OpenSSL 1.0.2 detection

Upstream ticket: https://freeswitch.org/jira/browse/FS-7314

It is resolved upstream :
https://freeswitch.org/stash/projects/FS/repos/freeswitch/commits/320d2eae5253f2a5c7f440e9291d4cc3fab7e3d2

BlueMax:
I get a similar error like in the referenced issue #630 above (after including the upstream patch):
> Checking for Networking support...
checking for in_addr in netinet/in.h... yes
checking if fd == socket on this platform... yes
checking style of gethostbyname_r routine... glibc2
checking 3rd argument to the gethostbyname_r routines... char
checking if TCP_NODELAY setting is inherited from listening sockets... yes
checking if O_NONBLOCK setting is inherited from listening sockets... yes
checking whether TCP_NODELAY and TCP_CORK can both be enabled... configure: error: in `/home/bluemax/openwrt/trunk/build_dir/target-i386_i486_uClibc-0.9.33.2/freeswitch-1.5.15b/libs/apr':
configure: error: cannot run test program while cross compiling
See `config.log' for more details
configure: error: ./configure.gnu failed for libs/apr
make[2]: *** [/home/bluemax/openwrt/trunk/build_dir/target-i386_i486_uClibc-0.9.33.2/freeswitch-1.5.15b/.configured_yyyyyynynynnnynnyynynnyyynynynnnnnnnnnnnnnnnnnnyynnnnnnnynnnynnnnnynnnnnnnnnnnnnnynnnnnnnnnynnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnynnynnnnnnnnnnnynynnnnnnnnnnnnnnnyyyyyyyyy] Error 1
make[2]: Target `compile' not remade because of errors.
make[2]: Leaving directory `/home/bluemax/openwrt/trunk/feeds/telephony/net/freeswitch'
make[1]: *** [package/feeds/telephony/freeswitch/compile] Error 2
make[1]: Leaving directory `/home/bluemax/openwrt/trunk'
make: *** [package/feeds/telephony/freeswitch/compile] Error 2

[...]

PR #250, opened 2018-01-21
freeswitch: disable features depending on missing packages

hnyman:
I wonder if the whole old freeswitch should be removed, as it currently causes config warnings. All other packages have been corrected in the last few days, but freeswitch has several broken dependencies, either to packages still in oldpackages or totally missing.

```
perus@ub1710:/OwrtLEDE/r7800$ make defconfig
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libsqlite2', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'flite', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libyuv', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'hiredis', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libilbc', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libmemcached', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libsilk', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'libg7221', which does not exist
WARNING: Makefile 'package/feeds/telephony/freeswitch/Makefile' has a dependency on 'freeradius-client', which does not exist
#
# configuration written to .config
#
```

I made a stab in cleaning those dependencies out of the Makefile, and this PR contains those changes:

> Disable features that depend on packages that either are in the oldpackages repo or are totally missing. Recent metadata handling changes generate cause warnings about dependencies to non-existing packages. Most of the features have been marked BROKEN for ages.
> * sqlite2: remove as sqlite3 is supported
> * flite: in oldpackages
> * libyuv, fsv: unsupported
> * hiredis: unsuported
> * libilbc: unsupported
> * libmemcached: unsupported
> * libsilk: unsupported
> * libg7221, siren: unsupported
> * radius-cdr, xml-radius: freeradius-client is unsupported

However, when compile-testing it with ar71xx, the package still fails to same compile error due that has been discussed in https://github.com/openwrt/telephony/issues/10#issuecomment-262698094 
> checking for robust cross-process mutex support... configure: error:  ... /libs/apr': configure: error: cannot run test program while cross compiling 

(also #35 looks similar)

So, I wonder if it would be easier to deprecate the whole freeswitch package, as freeswitch-stable is apparently the cleaned-up replacement. I think that freeswitch has not compiled in buildbot for ages, so the package is rather deprecated in reality.

In any case, the changes in this PR make the config warnings go away.

@micmac1 @jow-

### DIFF
--- a/net/freeswitch/Config.in
+++ b/net/freeswitch/Config.in
@@ -77,11 +77,6 @@ menu "Configuration"
       help
 	Compile libs/apr-util with PostgreSQL support.
 
-    config FS_WITH_SQLITE2
-      bool "SQLITE2"
-      help
-	Compile libs/apr-util with SQLITE2 support.
-
     config FS_WITH_SQLITE3
       bool "SQLITE3"
       help
@@ -160,12 +155,6 @@ menu "Configuration"
     default y
     help
 	Compile to build libs/libvpx.
-
-  config FS_WITH_LIBYUV
-    bool "Enable building libyuv"
-    default y
-    help
-	Compile to build libs/libyuv.
 
   config FS_WITH_LZMA
     bool "Enable liblzma usage in libtiff"
--- a/net/freeswitch/Makefile
+++ b/net/freeswitch/Makefile
@@ -17,6 +17,7 @@ PKG_NAME:=freeswitch
 PKG_SOURCE_PROTO:=git
 PKG_DISTNAME:=$(PKG_NAME)_$(PKG_SOURCE_PROTO)
 PKG_VERSION:=1.7.0
+PKG_RELEASE:=2
 
 
 #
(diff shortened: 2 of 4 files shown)


hnyman:
Looks like Mazi Lo visited the repo today and updated the package version, but didn't fix the missing packages config warning issues.

My PR is now in conflict :-(

hnyman:
I will close this, as mazilo followed the example and disabled the features linking to non-existing packages.

Sadly he forgot to disable "flite" that still gives the warning to all users, but hopefully that will get disabled by mazilo or some other telephony-repo committer sooner or later.

PR #253, opened 2018-01-26
bcg729: Remove BUILD_PATENTED.

neheb:
As of 2017, all patents have expired. Full announcement text below:

As of January 1, 2017 the patent terms of most Licensed Patents under the G.729 Consortium have expired.

With regard to the unexpired Licensed Copyrights and Licensed Patents of the G.729 Consortium Patent License Agreement, the Licensors of the G.729 Consortium, namely Orange SA, Nippon Telegraph and Telephone Corporation and Université de Sherbrooke (“Licensors”) have agreed to license the same under the existing terms on a royalty-free basis starting January 1, 2017.

For current Licensees of the G.729 Consortium Patent License Agreement, no reports and no payments will be due for Licensed Products Sold or otherwise distributed as of January 1, 2017.

### DIFF
--- a/libs/bcg729/Makefile
+++ b/libs/bcg729/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bcg729
 PKG_VERSION:=1.0.4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.linphone.org/releases/sources/bcg729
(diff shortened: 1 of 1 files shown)


micmac1:
Hi @neheb,

I think there's an argument to be made to keep this in place, because the general consensus seems to be that most patents have expired but not all. Check out the Wikipedia's G729 article, especially the Licensing paragraph. It says "As of January 1, 2017 the patent terms of most licensed patents under the G.729 Consortium have expired, the remaining unexpired patents are usable on a royalty-free basis."

From that point of view BUILD_PATENTED should stay in place in my opinion. It's nice that the remainder of patents can be used royalty-free. But BUILD_PATENTED has nothing to do with the royalties, it's solely a variable that says there is patent encumbered code in this package, which according to Wikipedia is still the case.

[G729 on Wikipedia](https://en.wikipedia.org/wiki/G.729)

Kind regards,
Seb

jslachta:
Hi @neheb and Sebastian @micmac1 ,

patents for g729a/b have expired. As far as I know the patents for the rest of the annexes are still valid. However bgc729 covers only annex A and annex B, so the BUILD_PATENTED dependency could be removed. I need to investigate a little bit more and then decide.

Thank you,

Jiri

PR #256, opened 2018-01-31
dahdi-linux: update source to snapshot

micmac1:
Maintainer: @jslachta 
Compile tested: arm_mpcore
Run tested: N/A

Description:
Hi Jiri,

on targets with kernel 4.14 dahdl-linux fails to build (snapshots), causing follow-up breakage. I suggest to temporarily use a git snapshot to work around this until upstream makes a new releases that includes the fixes.

I tried cherry-picking but couldn't get it to work. The snapshot build works, though (build-tested only).

Kind regards,
Seb


jslachta:
ACK

PR #257, opened 2018-02-03
freeswitch: disable flite module

hnyman:
Disable also flite module that depends on a non-existing package.

Mazilo fixed all other modules that had wrong dependencies, with be557c9bc07dd , but apparently forgot about flite. It still causes the config warning for all Openwrt/LEDE build system users.

Disable it as I suggested in #250 and mazilo did with be557c9bc07dd for other deps

Maintainer: mazilo


hnyman:
@jow- @micmac1 @jslachta 

Anybody with telephony commit rights interested enough, so that we get rid of the last remaining config WARNING spam related to freeswitch?

mazilo fixed all other missing package dependencies the same way with be557c9bc07d , but forgot this "flite". The whole option is marked BROKEN, so it is not even visbile for normal buildchain users. But still it causes the WARNING spam.

And as mazilo is not openly here in Github (which is strange for a commit access holder), it is really hard to pinpoint this to him.

so, please somebody just merge this.

jslachta:
Hi Hannu @hnyman,

I am merging this. Thank you.

PR #270, opened 2018-02-12
dahdi-linux: include linux/uaccess.h rather than asm/uaccess.h

micmac1:
Maintainer: @jslachta 
Compile tested: mips64_octeon, x86_64
Run tested: N/A

Description:
Hi Jiri,

found another build failure with newer kernels in dahdi-linux. Somebody already fixed the same for a different software, so I could just copy/paste, basically.

Kind regards,
Seb


micmac1:
Thanks Jiri!

PR #275, opened 2018-02-17
net/yate: some tiny updates

micmac1:
Maintainer: @jslachta 
Compile tested: x86_64, arc_archs, mips_24kc
Run tested: mips24kc

Description:
Hi Jiri,

Stumbled over an OpenWrt variable to disable FORTIFY_SOURCE. It's better to use this than removing CPPFLAGS.

Regards,
Seb


micmac1:
I think I'll put two more commits into this

jslachta:
Thank you for your hard work, Sebastian. Ack.

PR #301, opened 2018-03-13
Asterisk standard c++ lib

micmac1:
Maintainer: @jslachta 
Compile tested: x86_64, mips_24kc
Run tested: mips_24kc

Description:
Hi Jiri,

This removes the hardcoding of libstdcpp in the Asterisk and pjproject packages. If the preference is libuClibc++ it uses that instead. On OpenWrt libuClibc++ is default. It's about 1 MByte smaller on the router than libstdc++ and uses less RAM.

I've tested both 13 and 15, making phone calls both via chan_sip and pjsip. All OK.

The pull request also fixes some depends.

Kind regards,
Seb

### DIFF
--- a/libs/pjproject/Makefile
+++ b/libs/pjproject/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pjproject
 PKG_VERSION:=2.7.2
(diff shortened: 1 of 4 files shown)


jslachta:
Thank you, Seb.

PR #303, opened 2018-03-16
iksemel: use Google Storage API to get sources from former Googlecode

ryzhovau:
Maintainer: @jslachta
Compile tested: [Entware](https://github.com/Entware)
Run tested: Entware MIPSEL


jslachta:
ACK. Thank you, Alexander.

PR #283, opened 2018-02-24
Siproxd updates

micmac1:
Maintainer: @jslachta 
Compile tested: x86_64, arc, mips24kc
Run tested: mips24kc

Description:
Hi Jiri,

this is not related to FPIC. Just some stuff that caught my attention when checking siproxd.

Best regards,
Seb

PR #304, opened 2018-03-18
siproxd init code and usability improvements

guidosarducci:
Maintainer: @jslachta
Compile tested: ar71xx/mips24kc
Run tested: ar71xx/mips24kc under LEDE 17.01.4

Description:
Hi Jiri,

This PR includes the few changes discussed as part of #283 , including some UCI default config improvements suggested by @micmac1 .

The switchover to using callbacks to parse the UCI config file also exposed some 6-year-old related problems, which I've fixed and tested as part of https://github.com/openwrt/openwrt/pull/805. I'd appreciate if you could review that PR and comment as needed.

Note that the other PR is not a strict dependency for the current PR. The upstream callback problems result in spurious config file entries which, thankfully, siproxd ignores. However, those issues have likely caused historical problems for people making non-trivial use of callbacks, and should be fixed.

Kind regards,
Tony

### DIFF
--- a/net/siproxd/files/siproxd.config
+++ b/net/siproxd/files/siproxd.config
@@ -1,3 +1,25 @@
 config siproxd general
+	# Custom options allow using OpenWRT network names, and defaults should
+	# work out-of-the-box. If your SIP devices do not REGISTER externally,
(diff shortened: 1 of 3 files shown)


jslachta:
Thank you, Tony. Looks OK to me.

PR #310, opened 2018-04-08
Postgresql available now for ARC

micmac1:
Maintainer: @jslachta 
Compile tested: N/A
Run tested: N/A

Description:
Hello Jiri,

Since [this commit](https://github.com/openwrt/packages/commit/be07e031d04e5cab26856f1b5b5c2572f34ebb7b#diff-c3682966a20288b14bbcd182a0c90569) postgresql compiles for ARC as well. So we can remove the '@!arc' depends from the related modules.

Kind regards,
Seb


jslachta:
ACK, thank you very much, Sebastian!

PR #320, opened 2018-04-15
asterisk-chan-lantiq: avoid VARIANTs

micmac1:
Maintainer: @jslachta 
Compile tested: mips_24kc
Run tested: N/A

Description:
Hi Jiri,

I'm thinking that maybe the build bots do not even attempt the build chan-lantiq because it's a VARIANT build. Maybe that's not supported during the nonshared builds.

It's just an idea, but do you mind if we try it out? Back in the day when I stripped chan-lantiq from the main asterisk packages, the first thing I did was set it up as a VARIANT package. So we didn't test this yet. Maybe it works (or maybe not) :-)

All the other nonshared packages do not use VARIANTs. So maybe, just maybe this is it.

Kind regards,
Seb


jslachta:
Lets try this out. 👍

PR #321, opened 2018-04-27
asterisk-chan-lantiq: fix dep for nonshared builds

micmac1:
Maintainer: @jslachta 
Compile tested: lantiq xway
Run tested: N/A

Description:
Hi Jiri,

I'm 99% sure now that I found the reason for the buildbots not building chan-lantiq. This commit should fix this.

I thought about reverting the previous commit, but the VARIANT version didn't look too nice due to the different git branches used upstream. I think the non-VARIANT Makefiles for chan-lantiq are easier on the eyes and should be kept.

The commit messages mentions the reason for the new sub menu. I didn't find a better way. Without it it would look really broken:

```
{M} asterisk13............................ Complete open source PBX, v13.19.2  --->                      │ │  
  │ │                                  <M> asterisk13-chan-lantiq............................. Lantiq channel driver (NEW)                      │ │  
  │ │                                  < > asterisk13-chan-mgcp........................................ MGCP support (NEW)                      │ │  
  │ │                                  < > asterisk13-chan-mobile......................... Bluetooth channel support (NEW)                      │ │  
  │ │                                  < > asterisk13-chan-motif............................. Jingle channel support (NEW)                      │ │  
  │ │                                  < > asterisk13-chan-ooh323............................. H.323 channel support (NEW)
```

Kind regards,
Seb

### DIFF
--- a/net/asterisk-13.x-chan-lantiq/Makefile
+++ b/net/asterisk-13.x-chan-lantiq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=asterisk13-chan-lantiq
 PKG_VERSION:=20180215
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/kochstefan/asterisk_channel_lantiq.git
@@ -27,12 +27,12 @@ PKG_FLAGS:=nonshared
 include $(INCLUDE_DIR)/package.mk
 
 define Package/$(PKG_NAME)
-  SUBMENU:=Telephony
+  SUBMENU:=Telephony Lantiq
   SECTION:=net
   CATEGORY:=Network
   TITLE:=Lantiq channel driver
   URL:=https://github.com/kochstefan/asterisk_channel_lantiq
-  DEPENDS:=asterisk13 +kmod-ltq-vmmc
+  DEPENDS:=+asterisk13 +kmod-ltq-vmmc
 endef
 
 define Package/$(PKG_NAME)/description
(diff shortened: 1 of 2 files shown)


jow-:
@micmac1 @jslachta - if I see this correctly, the change will put a higher burden on the phase 1 builders due to the amount of dependencies required for asterisk to build. 

Is there any chance to optimize the asterisk Makefiles to only do the minimum required thing to provide whatever headers are needed for chan-lantiq?

I did a similar change for openvswitch in the past to not require python to build when only kmod-openvswitch is needed: https://github.com/openwrt/packages/commit/bf7ef2af120914598aa9b177ab0960a605c214f0

micmac1:
Hi @jow- and @jslachta 

I looked into this. The first header chan-lantiq includes is "asterisk.h". That header includes "asterisk/autoconfig.h", which is auto-generated during the configure phase.

I checked how many depends are actually added by comparing to the falcon build, which should be more or less the same as far as pulled-in packages go:

jansson
libcap
libedit
libopenssl
libsqlite3

So these are added by asterisk to the packages that get compiled anyway (ncurses, zlib etc.). The problem is that if you drop these depends then configure fails and "asterisk/autoconfig.h" doesn't get generated.

OK, maybe we could skip the compile stage then and just install the headers. But that would complicate the Makefile quite a bit. Is that really worth it? We have loads of extra modules with Asterisk. None of them is selected during the lantiq builds. We have the "AST_ENABLE" feature in the Makefile that prevents the build of every module and every utility that is not selected. So just the base stuff is built.

Additionally ccache is used, both for building menuselect and asterisk itself. On a ca. 5 year old laptop with HDD and i3-3110M I get this:

```
sk@hotdamn ~/tmp/openwrt $ time make -j1 package/asterisk-13.x/{clean,compile} V=s &> ~/compile-ast13-1.txt

real	0m55,621s
user	0m35,693s
sys	0m4,170s
sk@hotdamn ~/tmp/openwrt $
```

Doesn't look too bad. Also, chan-lantiq will only get selected during builds where "kmod-ltq-vmmc" is available. And that modules is only available on xway and xway200. So chan-lantiq is not build for every target, just these two.

Another option would be to prevent the build again in trunk (by reverting this commit) and just build chan-lantiq in the release branches.  So they would only be build occasionally for the releases, but not during the snapshot compiles. I think that would be fair still. Personally I'd prefer that to complicating the Makefile any more.

Please let me know what you guys want done and I'll send a PR for it.

Best regards,
Seb

PR #329, opened 2018-05-20
siproxd: fix init code when multiple sections are defined

mkuron:
Maintainer: @jslachta
Compile tested: LEDE 17.01.4 x86-generic
Run tested: LEDE 17.01.4 x86-generic

Description:

When multiple sections were defined in uci for siproxd, all but the last one didn't result in a running siproxd process. The configuration file was written incompletely for all but the last section defined. The service was only started for that configuration file. It seems like that issue was introduced by @guidosarducci in https://github.com/openwrt/telephony/pull/304. Since that patch went into the lede-17.01 and openwrt-18.06 branches, this one should be cherry-picked too.

### DIFF
--- a/net/siproxd/files/siproxd.init
+++ b/net/siproxd/files/siproxd.init
@@ -95,6 +95,7 @@ siproxd_cb() {
 			# Initialize section processing and save section name.
 			"siproxd")
 				sec="$2"
+				secs="$secs $sec"
 				if [ -f "$siproxd_conf_prefix$sec.conf" ]; then
 					rm "$siproxd_conf_prefix$sec.conf"
 				fi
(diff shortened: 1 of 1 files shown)


guidosarducci:
Thanks very much for spotting this -- multiple configs are something I test but must have missed it this time. Having said that, I think the workaround here defeats the purpose of callbacks and so isn't ideal.

I do have a more recent update to support procd and restructure the code, which fixes this bug as a side-effect. Would you mind reviewing and trying out #332 instead and sharing any feedback?

Glad to know someone else actually uses `siproxd`!

mkuron:
@guidosarducci, thanks for the reply. I knew my patch wasn’t ideal; I just wanted to offer an actual fix instead of only reporting the issue. I‘m happy to give #332 a try and will report back soon.

mkuron:
#332 does fix the issue for me, so I am closing this pull request in its favor.

guidosarducci:
Good to hear, thank you for testing that.

PR #338, opened 2018-06-10
freeswitch: fix a memory leak with openssl 1.1.0

cotequeiroz:
Maintainer: Mazi Lo (I couldn't find his username)
Compile tested: brcm47xx, openwrt master
Run tested: none

Description:
Fixed a memory leak when compiled with openssl 1.1.0.  It should be very simple to see this.  CTX is allocated, but it is not being freed, but only reset (cleanup is defined to reset in openssl 1.1.0).  I caught this while backporting openssl 1.1.0 support to v. 1.6.20, and submitted the patch upstream.

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>


jslachta:
Hi Eneas @cotequeiroz ,

please also bump PKG_RELEASE along with your changes. Thank you!

cotequeiroz:
I've created an issue for this upstream:
https://freeswitch.org/jira/browse/FS-11193

PR #344, opened 2018-06-26
(for OpenWrt 18.06): asterik: fix AST-2018-007 and AST-2018-008

micmac1:
Maintainer: @jslachta 
Compile tested: ar71xxx
Run tested: N/A

Description:
Add patches for security advisories.


jslachta:
Thank you, seb!

PR #346, opened 2018-07-08
asterisk: fix iconv related build issue

micmac1:
Maintainer: @jslachta 
Compile tested: archs
Run tested: don't have archs hardware

Description:
Hello Jiri,

I'm cleaning up after the fallout from the libmariadb 10.2 upgrade. Asterisk is affected and needs a bit of tweaking for the uclibc compiles.

Kind regards,
Seb


jow-:
Same as for Yate, this should really be an `include $(INCLUDE_DIR)/nls.mk` instead

micmac1:
Pull request updated.

PR #347, opened 2018-07-08
yate: fix uclibc build issue

micmac1:
libmariadb 10.2 needs to be linked in together with iconv. On musl and
glibc iconv is part of libc. That's not the case for uclibc, where
libiconv-full needs to be used. This commit aids yate-mod-mysqldb in
finding libiconv.

Also fixes a typo in PKG_CONFIG_DEPENDS.

Maintainer: @jslachta 
Compile tested: archs
Run tested: don't have archs hw

Description:
Hello Jiri,

Same as for asterisk. I think the other packages in telephony are/will be fine.

Best regards,
Seb


jow-:
Please encapsulate that logic in `include/nls.mk` it is meant to handle such issues.

micmac1:
Hi Jo,

I'm not particularly fond of this idea as then libiconv-full is only available if BUILD_NLS is set - which is not the default. So if people wanted to use the mysql module they'd need to build it manually.

The way the commit sets it up it will be always available from the buildbots and install the libiconv-full depend if the mysql module is installed.

Kind regards,
Seb

jow-:
Why do we even need to depend on libiconv-full? Isn't the iconv stub enough?

jow-:
What I am getting at is that `iconv()` on musl is not really full-featured either but merely an API stub providing some charset conversions but certainly not the full range of glibc or GNU iconv. So if a reduced but API-complete `iconv()` is enough on musl, then the OpenWrt/LEDE libiconv stub should be adequate for uclibc, in which case you should use the logic in `nls.mk` to align the behaviour across all libc variants.

micmac1:
Am July 9, 2018 12:11:52 PM UTC schrieb Jo-Philipp Wich <notifications@github.com>:
>What I am getting at is that `iconv()` on musl is not really
>full-featured either but merely an API stub providing some charset
>conversions but certainly not the full range of glibc or GNU iconv. So
>if a reduced but API-complete `iconv()` is enough on musl, then the
>OpenWrt/LEDE libiconv stub should be adequate for uclibc, in which case
>you should use the logic in `nls.mk` to align the behaviour across all
>libc variants.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/openwrt/telephony/pull/347#issuecomment-403457822

Hi Jo,

Did not know that. Thanks for explaining. Will try it out with mariadb an some dependent packages.

Regards,
Seb

micmac1:
Hello @jow- 

I tried to find out what the expected/intended behaviour is when including nls.mk.

Question: Is the expected behaviour that when CONFIG_BUILD_NLS is set the binaries link to "$(STAGING_DIR)/usr/lib/libiconv-full/lib/libiconv.so" and when not they link to "$(STAGING_DIR)/usr/lib/libiconv-stub/lib/libiconv.a", no matter which libc is used?

For uclibc that seems to be the case (checked with libidn2). But it also seems the case for musl (did not try glibc, but I assume the outcome would be the same). So when nls.mk is included and musl is used, the musl-internal iconv will not be used? That is intended?

Thanks in advance!

Kind regards,
Seb

micmac1:
Hi all!

First off, hello @jow-. I found that mocp behaves like libidb2 wrt iconv. No matter the libc, iconv-stub is used when BUILD_NLS is off, iconv-full otherwise. So libc's iconv isn't used. I updated mariadb to mimmick this behavior. And then just including nls.mk is enough. Again, thanks.

Hello @jslachta I've updated the pull request. I added an upgrade to yate 6.1.0. @cotequeiroz mentioned it's available and that it takes care of the openssl 1.1.0 compatibility.

Kind regards,
Seb

cotequeiroz:
I suggest you change the title of this to reflect the upgrade.
Cheers

micmac1:
Am July 12, 2018 9:00:26 PM UTC schrieb Eneas Queiroz <notifications@github.com>:
>I suggest you change the title of this to reflect the upgrade.
>Cheers
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/openwrt/telephony/pull/347#issuecomment-404649190

Hi Eneas,

Will do ;-) Thx


Seb

cotequeiroz:
Are we good to go here?  This is one of the last packages needing changes (the version bump would be enough) before we can apply openssl 1.1.0.
Cheers.

cotequeiroz:
Hi Seb
Thanks.

PR #356, opened 2018-08-01
freeswitch-stable: fix memory leak in mod_event_multicast

cotequeiroz:
Maintainer: @micmac1 
Compile tested: ramips/brcm47xx, openwrt master
Run tested: none

Description:
Applied a patch, submitted upstream at https://freeswitch.org/jira/browse/FS-11193
that fixes two memory leaks in `mod_event_multicast.c`.

In `src/mod/event_handlers/mod_event_multicast/mod_event_multicast.c`, `event_handler()` function, when encrypting data using openssl 1.1.0, the context ctx is allocated by `EVP_CIPHER_CTX_new()`, then data is encrypted, and at the end, EVP_CIPHER_CTX_cleanup is called. This function resets the context so that it can be used again, but ctx is not freed. The correct call would be to `EVP_CIPHER_CTX_free()`, which frees the memory.
The same thing happens in the decryption function: `ctx` is allocated, but never freed. 

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>

### DIFF
--- a/net/freeswitch-stable/Makefile
+++ b/net/freeswitch-stable/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PRG_NAME:=freeswitch
 PKG_NAME:=$(PRG_NAME)-stable
 PKG_VERSION:=1.8.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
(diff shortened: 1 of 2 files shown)


micmac1:
LGTM

Thanks Eneas!

PR #377, opened 2018-09-28
asterisk related updates

dhewg:
Maintainer: @jslachta 
Compile tested: sunxi, master
Run tested: sunxi, master

Description:
This updates both asterisk versions to the current release, prepares them for the current pjproject 2.8 (using upstream patches) and then switches over to that version.

asterisk-13.x was only compile time tested, while asterisk-15.x was updated on my sunxi cubieboard2 setup. That works without issues so far (tested both pjproject versions).

Additionally, the open opus codec is added to asterisk-15.x. This is the same project as can be found in e.g. debian [0]. Tested successfully with linphone and transcoding in each direction.

Thanks!
Andre

[0] https://packages.qa.debian.org/a/asterisk-opus.html


micmac1:
I think we should keep pjsip 2.7 for now and just wait until asterisk upstream provides official releases which support 2.8 out of the box. That also lets us deal easily with potential future pjsip vulnerabilities (we just track asterisk upstream).

Adding opus support would be nice OTOH.

Just my 2 cents.

micmac1:
Hi!


I'm not aware of any unpatched vulns in owrt's pjproject. I just meant that I don't see any upside in carrying the proposed extra patches and upgrading to 2.8 right now.


Kind regards
Seb

dhewg:
I dropped the patches wrt to pjproject 2.8 and added one that updates pjproject's configuration so it matches with upstream asterisk

net/asterisk-15.x/Makefile
```diff
@@ -840,6 +849,7 @@ $(eval $(call BuildAsterisk15Module,codec-g726,Signed linear to G.726 translatio
 $(eval $(call BuildAsterisk15Module,codec-gsm,linear to GSM translation,translate between signed linear and GSM,,,codec_gsm,,))
 $(eval $(call BuildAsterisk15Module,codec-ilbc,linear to ILBC translation,translate between signed linear and ILBC,,,codec_ilbc,,))
 $(eval $(call BuildAsterisk15Module,codec-lpc10,Linear to LPC10 translation,translate between signed linear and LPC10,,,codec_lpc10,,))
+$(eval $(call BuildAsterisk15Module,codec-opus-open-source,OPUS codec,Translate between signed linear and Opus (Open Codec),+libopus,,codec_opus_open_source,,))
```
micmac1: Why not also add format_ogg_opus_open_source and format_vp8? Debian packages them as well, see https://packages.debian.org/stretch/amd64/asterisk-opus/filelist

net/asterisk-13.x/Makefile
```diff
@@ -532,7 +532,6 @@ CONFIGURE_ARGS+= \
 	--without-nbs \
 	--with-libedit="$(STAGING_DIR)/usr" \
 	--with-libxml2 \
-	--with-ncurses="$(STAGING_DIR)/usr" \
```
micmac1: ncurses is not dropped from depend like commit message suggests.

libs/pjproject/patches/150-config_site.patch
```diff
@@ -11,34 +11,23 @@
 + * Since both pjproject and asterisk source files will include config_site.h,
 + * we need to make sure that only pjproject source files include asterisk_malloc_debug.h.
 + */
-+
-+/* #if defined(MALLOC_DEBUG) && !defined(_ASTERISK_ASTMM_H)
-+ * #include "asterisk_malloc_debug.h"
-+ * #endif
-+ */
++#if 0 && !defined(_ASTERISK_ASTMM_H)
```
micmac1: This shouldn't get in. We compile pjproject standalone, not from within Asterisk, so it makes not sense to have something in this file that references asterisk sources (even if it'll not get used anyway).

libs/pjproject/patches/150-config_site.patch
```diff
@@ -11,34 +11,23 @@
...
++#include "asterisk_malloc_debug.h"
++#endif
 +
 +/*
 + * Defining PJMEDIA_HAS_SRTP to 0 does NOT disable Asterisk's ability to use srtp.
 + * It only disables the pjmedia srtp transport which Asterisk doesn't use.
 + * The reason for the disable is that while Asterisk works fine with older libsrtp
 + * versions, newer versions of pjproject won't compile with them.
 + */
-+
```
micmac1: Why delete this comment? Is it not valid anymore?

libs/pjproject/patches/150-config_site.patch
```diff
@@ -11,34 +11,23 @@
...
-+ * libsrtp, otherwise pjsip would just build the internal one.
-+ */
-+
 +#define PJMEDIA_HAS_SRTP 0
 +
-+/*
-+ * Defining PJMEDIA_HAS_WEBRTC_AEC to 0 does NOT disable Asterisk's ability to use
-+ * webrtc.  It only disables the pjmedia webrtc transport which Asterisk doesn't use.
-+ */
-+#define PJMEDIA_HAS_WEBRTC_AEC 0
```
micmac1: Why remove the define? Is asterisk now using pjmedia's webrtc transport?

libs/pjproject/patches/150-config_site.patch
```diff
@@ -11,34 +11,23 @@
...
-+
 +#define PJ_HAS_IPV6 1
++#if !defined(AST_DEVMODE) && !defined(PJPROJECT_BUNDLED_ASSERTIONS)
```
micmac1: This should not be in here, this is standalone pjsip, not from within asterisk. AST_DEVMODE shouldn't be referenced.

libs/pjproject/patches/150-config_site.patch
dhewg: The patch was created from the [config_site.h file from asterisk](https://github.com/asterisk/asterisk/blob/master/third-party/pjproject/patches/config_site.h). All changed to this patch are because of asterisk commits, see [here](https://github.com/asterisk/asterisk/commits/master/third-party/pjproject/patches/config_site.h).
The only change I did was that 'if 0' to not use the asterisk header (it was already there, just commented), because I tried to make the diff against the upsteam file as minimal as possible. That'll make future updates easier.

libs/pjproject/patches/150-config_site.patch
dhewg: Its a preprocessor define, nothing is referenced. Since standalone pjproject doesn't know anything about AST_DEVMODE the compile won't ever see that block.

net/asterisk-15.x/Makefile
micmac1: Hello Andre, makes sense, thanks for explaining.

libs/pjproject/patches/150-config_site.patch
micmac1: OK, this can be removed because we pass --disable-libwebrtc.

libs/pjproject/patches/150-config_site.patch
micmac1: But why the extra clutter? Let's just keep NDEBUG unconditionally.

micmac1:
Another thought, though. The opus codec in here is just for 15. I think it would be cool to add it to 13 as well. That is to say that it would be nice to have it in all versions. But if we patch Asterisk itself then obviously we'd need to duplicate code. So I think it would be better if we separated the opus codec into it's own package (same as we do for bcg729 for instance) and then use VARIANTs.

I'll whip something up and post a pull request.

dhewg:
Patches updated:
* ncurses dependency for 13.x really dropped
* removed clutter from config.j

dhewg:
Oh and I removed the opus patch in favor of your PR

micmac1:
Hello Andre,

Thank you! Looks good to me.

Kind regards,
Seb

PR #379, opened 2018-10-02
yate: Revert to internal regex implementation

### LINKED ISSUE #378: yate: number regex routing fails
Maintainer: @jslachta
Environment: ar71xx, dir-825 b1, OpenWrt 18.06.1

Description:
I've been using regexroute.conf to do call routing in yate. I've had entries like this:
```
^12345$=fork ...
```
Then when a call for number 12345 is received it's forked to the phones I've specified.
This has worked well with lede 17.01.4, but stopped working when I upgraded to OpenWrt 18.06.1.

I've done some debugging and found that this:
```
^.*$=echo \0
```
.. prints "sip:12345", which seems to indicate that the "sip:" prefix is now included in the routing string.

When I ran lede 17.01.4 I compiled and used yate 6.0.0 successfully so I think that yate hasn't changed (OpenWrt 18.06.1 also includes yate 6.0.0). I've looked through the OpenWrt patches added the past year, but they don't look related to this problem.

So, I get a feeling that something in the yate environment has changed (musl or libraries?).

Has anyone noticed any similar behaviour?

robho:
.. because the musl implementation doesn't seem to be fully compatible
with yate. We switched to the musl implementation in f6ad95d.

Yate has this regexp:
^([[:alpha:]][[:alnum:]]+:)?/?/?([^[:space:][:cntrl:]@]+@)?([[:alnum:]._+-]+|[[][[:xdigit:].:]+[]])(:[0-9]+)?

Given a string like

  sip:012345678@11.111.11.111:5060;user=phone

musl's regexec() returns these matches:
```
index start end
0     -1    0
1     0     32   sip:012345678@11.111.11.111:5060
2     -1    -1
3     0     14   sip:012345678@
4     14    27   11.111.11.111
5     27    32   :5060
```
.. but this is what yate expects:
```
index start end
0     -1    0
1     0     32   sip:012345678@11.111.11.111:5060
2     0     4    sip:
3     4     14   012345678@
4     14    27   11.111.11.111
5     27    32   :5060
```
Fixes #378

Maintainer: @jslachta 
Compile tested: ar71xx, dir-825 b1, OpenWrt 18.06.1
Run tested: ar71xx, dir-825 b1, OpenWrt 18.06.1 - call routing now works as documented

### DIFF
--- a/net/yate/Makefile
+++ b/net/yate/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=yate
 PKG_VERSION:=6.1.0-1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://yate.null.ro/tarballs/yate6/
@@ -124,6 +124,10 @@ CONFIGURE_ARGS+= \
 	--without-coredumper \
 	--without-doxygen \
 	--without-kdoc
+
+# The regexp implementation of musl 1.1.19 is not fully compatible with yate
+CONFIGURE_ARGS+= \
+	--enable-internalregex
 
 ifneq ($(CONFIG_PACKAGE_$(PKG_NAME)-mod-isaccodec),)
 CONFIGURE_ARGS+=$(if $(CONFIG_SOFT_FLOAT),--disable-isac-float --enable-isac-fixed,--disable-isac-fixed --enable-isac-float)


robho:
Thanks for taking a look at this Sebastian. I added the comment you suggested.

I think this fix is valuable to 18.06 branch also. Will/can you merge this PR to that branch also or do we need another PR for that branch?

micmac1:
On Tue, Oct 02, 2018 at 10:52:48AM -0700, Robert Högberg wrote:
> Thanks for taking a look at this Sebastian. I added the comment you
> suggested.
> 
> I think this fix is valuable to 18.06 branch also. Will/can you merge
> this PR to that branch also or do we need another PR for that branch?

Thanks for adding the comment. I agree that this should also go into
18.06. I'll raise a PR for 18.06 if nobody beats me to it. Jiri is the
maintainer, so he'll do the merging.

jslachta:
Thank you for taking a look at this issue. I acknowledge this PR.

PR #382, opened 2018-10-13
kamailio-5.x: make jsonrpc filter portable

micmac1:
With OpenWrt default awk - so awk from busybox - the filter is broken,
causing jsonrpc calls to fail, i.e. when running "kamctl ps". Below
patch makes the filter portable. Patch was already accepted upstream.

Maintainer: @jslachta 
Compile tested: mips24kc DIR-825 C1
Run tested: mips24kc DIR-825 C1

Description:

Hello Jiri,

This fixes kamctl runtime failures due to non-portable awk script.

Kind regards,
Seb


jslachta:
This must have somehow changed. It worked before. Thank you for the patch.

Issue #384, opened 2018-11-05
siproxd-mod-regex UCI list vs entry processing


biuklija:
-------------------------------

Maintainer: @guidosarducci 
Environment: OpenWrt 18.06.1 r7258-5eb055306f

Description: UCI processes siproxd regex entries as a list, they should be processed as entries

siproxd expects regex replace entries as follows: 
```
plugin_regex_desc    = A
plugin_regex_pattern = ^(sips?:)00
plugin_regex_replace = \1+

plugin_regex_desc    = B
plugin_regex_pattern = ^(sips?:)0
plugin_regex_replace = \1+385

plugin_regex_desc    = C
plugin_regex_pattern = ^(sips?:)
plugin_regex_replace = \1+38531
```

However, since there is no entry support for the plugins, the following blocks in /etc/config/siproxd:

```
list plugin_regex_desc  A
list plugin_regex_pattern ^(sips?:)00
list plugin_regex_replace \1+

list plugin_regex_desc B
list plugin_regex_pattern ^(sips?:)0
list plugin_regex_replace \1+385

list plugin_regex_desc  C
list plugin_regex_pattern  ^(sips?:)
list plugin_regex_replace \1+38531	
```
get sorted in /var/etc/siproxd/siproxd-general.conf

```
plugin_regex_desc = A B C
plugin_regex_pattern = ^(sips?:)00
plugin_regex_pattern = ^(sips?:)0
plugin_regex_pattern = ^(sips?:)
plugin_regex_replace = 1+
plugin_regex_replace = 1+385
plugin_regex_replace = 1+38531
```

micmac1:
Hello Ivan,

I think this is a long-standing issue. I tried oldest init script in github from commit 100f27f3bb693522e8f2617f865e6a13aa406cb4 and it wrote
```
plugin_regex_desc = A
plugin_regex_desc = B
plugin_regex_desc = C
plugin_regex_pattern = ^(sips?:)00
plugin_regex_pattern = ^(sips?:)0
plugin_regex_pattern = ^(sips?:)
plugin_regex_replace = \1+
plugin_regex_replace = \1+385
plugin_regex_replace = \1+38531
```
into /var/etc/siproxd-general. With current init script I get the same in in /var/etc/siproxd/siproxd-general.conf.

When I trace the init script it's visible that

`config_load 'siproxd'`

is causing the wrong order.

```
config siproxd '"'"'general'"'"'              
        option interface_inbound '"'"'lan'"'"'
[...]

PR #387, opened 2018-11-07
siproxd: add link to Wiki page

micmac1:
Add some examples for the regex plugin to UCI configuration. Examples
provided by @biuklija in issue #384.

Maintainer: @jslachta 
Compile tested: No (just added comments to config file)
Run tested: No

Description:
Fixes issue #384 

Kind regards,
Seb

### DIFF
--- a/net/siproxd/Makefile
+++ b/net/siproxd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=siproxd
 PKG_VERSION:=0.8.2
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=@SF/siproxd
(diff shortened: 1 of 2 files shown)


guidosarducci:
Hi Seb,

Although I first added a bunch of these commented-out usage examples to the UCI config, I've started to have "committers remorse" as I realize the file becomes more and more opaque.

I'm thinking now it might be better to pull basic usage information and examples into a co-located "README.md" file, and keep the UCI config as a minimal working config.

This seems more commonly done on the other OpenWrt repos, where it's also been pointed out that any config file comments would be discarded during UCI file processing/rewriting.

What are your thoughts?

Best regards,
Tony

guidosarducci:
Hi Seb,

Thanks for the reminder about (b)! I had forgotten about the absence of any wiki articles, despite being annoyed about it long ago. That seems like a more widely useful option in the long term.

Given an "open" timeline for doing wiki updates though, I've no problem if you want to still merge the examples and give them a temporary home in the meantime.

Thanks again,
Tony

guidosarducci:
Hi Ivan,

Thanks for the offer to contribute; that would be very helpful since not everyone has experience among all the plugins. I've can add a wiki to my (slightly long) list of OpenWrt TODOs, but please don't time me since this would be my first wiki article!

The official docs are a bit lacking, so I expect Thomas Ries is fairly swamped, but he has been very responsive when I've tried to upstream patches in the past.

Like you, I also had to look through source code sometimes while I was updating this package, but the best up-to-date resource I found was the "hidden" examples file in the source tarballs:
`siproxd-0.8.3dev-24Mar2018/doc/siproxd.conf.example`

All the best,
Tony

micmac1:
Hello Tony and Ivan,

The good thing is that the wiki article doesn't need to be complete from the start. It can grow naturally over time.

I think for a start a short description copied from upstream will do. Maybe plaster the regex plugin comments afterwards.

Can we agree on the document location voip/sipproxd? Then I'll just start with the above.

Kind regards,
Seb

guidosarducci:
Yes, please. And thanks!

micmac1:
OK, so I added a Wiki page [here](https://openwrt.org/docs/guide-user/services/voip/siproxd) and changed the commit of this pull request to just link there.

I think this is OK like this. As said, the wiki page can be extended any time.

Kind regards,
Seb

guidosarducci:
Hi Seb,

Pardon the delay. And thanks, I saw the wiki page and that makes a great start for us to build upon!

I also think it would be preferable to add the wiki link to the Makefile package description, as done in many other packages: e.g.
```
define Package/siproxd/description
 Siproxd is a proxy/masquerading daemon for the SIP protocol. Refer to https://openwrt.org/docs/guide-user/services/voip/siproxd for configuration details and examples.
endef
```
This has the advantage of being available in the package catalog, and hence anytime both pre- and post-installation for someone looking for information, using only `opkg info siproxd`. And given the above, I recommend also removing any overloading of `postinstall` for user messaging. That doesn't seem an appropriate usage for the scriptlet and more complexity than we need.

Warm regards,
Tony

micmac1:
Hi Tony,

Good idea. I'll close this pull request and the maybe you can add one for that when you get around to it. ;-)

Kind regards
Seb

PR #388, opened 2018-11-10
baresip: remove recursive depend on libx264

micmac1:
Recursive depend was mentioned by @thess. This commit does away with it.

Maintainer: @jslachta 
Compile tested: mvebu
Run tested: N/A

Description:
Hi all,

This is to fix a recursive depend.

Regards,
Seb

### DIFF
--- a/net/baresip/Makefile
+++ b/net/baresip/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=baresip
 PKG_VERSION:=0.5.9
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=http://www.creytiv.com/pub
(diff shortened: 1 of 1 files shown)


micmac1:
Hello @thess 

Thanks for your feedback! I'll just close this pull request, then.

At first I though I should change it, maybe introduce a config switch and make the depend on libx264 optional. My idea was that then mod-avcodec could be built even when BUILD_PATENTED is off. But then I realized that ffmpeg-full will only be built when BUILD_PATENTED is on :-) Hehe. Man, what a dilemma.

Anyway, thanks!

Have a good Sunday, Ted!

Best regards,
Seb

alfredh:
hi guys,

the baresip avcodec module has 2 ways of using the H.264 encoder:

1. use libx264 directly
2. use ffmpeg libavcodec with the enabled H.264 encoders (such as libx264).

method 2 provides much more flexibility with regards to using e.g.
hardware accelerated codecs.

in the future we might even drop support for (1) and only use (2).

if usage of libx264 is cause problems for OpenWRT I suggest that you
always compile baresip avcodec with `USE_X264=` (i.e. empty/disabled).


Alfred

micmac1:
Am November 11, 2018 9:46:25 AM UTC schrieb "Alfred E. Heggestad" <notifications@github.com>:
>hi guys,
>
>the baresip avcodec module has 2 ways of using the H.264 encoder:
>
>1. use libx264 directly
>2. use ffmpeg libavcodec with the enabled H.264 encoders (such as
>libx264).
>
>method 2 provides much more flexibility with regards to using e.g.
>hardware accelerated codecs.
>
>in the future we might even drop support for (1) and only use (2).
>
>if usage of libx264 is cause problems for OpenWRT I suggest that you
>always compile baresip avcodec with `USE_X264=` (i.e. empty/disabled).
>
>
>Alfred
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/openwrt/telephony/pull/388#issuecomment-437656241

Hello Alfred, so nice of you to chime in. I'll raise a pull request to disable 1).

Kind regards,
Seb

thess:
As of now since the default build environment (buildbots) have BUILD_PATENTED off, nothing will have libx264 in it on our downloads. Only custom builds can have libx264 built if BUILD_PATENTED is selected. In both soft & hard float cases, libx264 will be included with libffmpeg-full if it is selected.

Sorry for all the confusion, I had BUILD_PATENTED on in my builds and didn't remember it was turned off earlier this year. I think the ffmpeg craziness is stable for a bit.

PR #395, opened 2018-12-09
siproxd: add wiki link and fix debug log capture

guidosarducci:
**Maintainer**: @jslachta 
**Compile tested**: ar71xx, DIR-835, LEDE-17.01.5
**Run tested**: (as above) Confirmed wiki link in package metadata and capture of debug output in syslog
**CC**: @micmac1 

**Description**:
Hi Jiri, Seb,
This is a follow-up from the examples/docs/wiki discussions in https://github.com/openwrt/telephony/issues/384 and https://github.com/openwrt/telephony/pull/387, and adds a package-metadata link to the OpenWrt wiki for `siproxd`.  While troubleshooting #384 I also noticed debug output wasn't being captured to syslog, so also fixed that. Please have a look...

Best regards,
Tony

### DIFF
--- a/net/siproxd/Makefile
+++ b/net/siproxd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=siproxd
 PKG_VERSION:=0.8.2
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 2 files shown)


jslachta:
Ack.

Thank you, Tony.

PR #401, opened 2018-12-26
asterisk-chan-sccp: update to 4.3.1 snapshot

micmac1:
Maintainer: @jslachta 
Compile tested: ath79
Run tested: N/A

Description:
Update chan-sccp.


micmac1:
Load-tested, like previously, as I don't sccp hardware. Loads fine :)

PR #402, opened 2018-12-26
asterisk-g72x: update to 1.4.3

micmac1:
Maintainer: @jslachta 
Compile tested: ath79
Run tested: N/A

Description:
Update to 1.4.3, includes asterisk 16 support.


micmac1:
Run-tested on ath79. Transcoding between G729 and G711 between two extensions calling each other, both using SRTP.

PR #403, opened 2018-12-26
sipp: update to 3.5.2

micmac1:
Maintainer: @jslachta 
Compile tested: ath79
Run tested: N/A

Description:
Update to 3.5.2, drop patches.


micmac1:
Run-tested on ath79. I used the default uac scenario against an asterisk running on the same host. No problem.

PR #407, opened 2019-01-20
kamailio-5.x: bump to 5.2.1

micmac1:
Maintainer: @jslachta 
Compile tested: mips24kc
Run tested: mips24kc, dir-825-c1, tested phone calls via Kamailio using remote MariaDB server.

Description:
Hello Jiri,

It was a bit more tricky than expected. Anyway, works fine for me.

Kind regards,
Seb


micmac1:
Oh, the two patches I dropped are included upstream, so we don't need them anymore.

micmac1:
Added a comment to commit message regarding the two dropped patches.

jslachta:
Thank you, Sebastian. I also tested it and works as expected.

PR #408, opened 2019-01-27
dahdi-tools: rework patch

micmac1:
Currently builds for ARC are failing, causing follow-up breakage. It
seems that uclibc defines __GLIBC__ on the build bots.

This commit adds a test for execinfo.h to the configure script. The
result of the test is then used in xpp/xtalk/debug.c.

Maintainer: @jslachta 
Compile tested: mips24kc
Run tested: not relevant, build fix

Description:

Hello Jiri,

The patch I sent earlier doesn't seem to work very well. On the build bots the ARC builds are currently failing. So they must have __GLIBC__ defined there.

I reworked the patch to explicitly check for the execinfo header. This should fix it.

Kind regards,
Seb

### DIFF
--- a/libs/dahdi-tools/Makefile
+++ b/libs/dahdi-tools/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dahdi-tools
 PKG_VERSION:=3.0.0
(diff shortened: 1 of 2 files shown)


jslachta:
Thank you, Seb!

PR #411, opened 2019-03-02
asterisk-16: new package

micmac1:
Maintainer: @jslachta 
Compile tested: master, ath79 (musl) + arc (uclibc)
Run tested: ath79 master

Description:
Hello all,

This is not ready for merging yet. There is a problem with segmentation faults when the module loader starts. I've had help from the [musl list](https://www.openwall.com/lists/musl/2019/02/28/4) and raised as issue for it [upstream](https://issues.asterisk.org/jira/browse/ASTERISK-28319).

According to [asterisk wiki](https://issues.asterisk.org/jira/browse/ASTERISK-28319) Asterisk 15 will be EOL 2019-10-03, so getting Asterisk 16 in would be nice. Not to mention that 16 is an LTS release. But before this can be merged something needs to be done regarding the segmentation fault.

Kind regards,
Seb

PR #412, opened 2019-03-05
Asterisk 16 additions

dhewg:
Maintainer: @jslachta 
Compile tested: arm/master
Run tested: arm/master

Description:
It seems asterisk gets no testing at all with musl, so I had to add two patches to get it working in the first place. But this exact version submitted here is running on my openwrt cubieboard2 for a few weeks now - and it works for me.

With this new package I chose to not add all the deprecated modules. With the idea that we can add those if there's demand for it.

* Add two pjproject patches, taken from the asterisk 16 tree
* Add the new v16 LTS release
* Add asterisk 16 variants of g72x and opus

Thanks!

### DIFF
--- a/libs/pjproject/Makefile
+++ b/libs/pjproject/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pjproject
 PKG_VERSION:=2.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=pjproject-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=http://www.pjsip.org/release/$(PKG_VERSION)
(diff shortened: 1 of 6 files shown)


dhewg:
Updated and removed the asterisk 16 port, @micmac1 takes care of it in #411. These patches go on top of it.

micmac1:
Hello @dhewg 

Checked the commits, looks all good to me. I'd squash the pjproject commits into one, but that'd only be a personal preference of mine.

Thanks for this!

Kind regards,
Seb

jslachta:
Hi @dhewg, would you be so kind to squash the pjproject commits? Thank you.

dhewg:
Squashed the pjproject commits, and added another patch which landed on the upstream asterisk 16 branch in the meantime

jslachta:
Thank you!

PR #416, opened 2019-04-13
kamailio-5.x: update hiredis patch

micmac1:
Maintainer: @jslachta 
Compile tested: ath79, master
Run tested: not needed, only compile-related

Description:
Hello Jiri,

neheb merged the libhiredis pc file fix yesterday, so the patch can be relieved of the workaround.

Kind regards,
Seb


jslachta:
Thank you, Sebastian.

PR #421, opened 2019-05-01
asterisk-16.x: update init to procd + hotplug

micmac1:
Converts the init script to procd. The user "asterisk" is added to the
system. This user will be used by default.

A hotplug script was also added. Both init and hotplug can be configured
in /etc/config/asterisk, via uci if desired.

Maintainer: @jslachta 
Compile tested: ath79
Run tested: ath79

Description:
Hi Jiri,

I still got three things that I was working on in the past few days. I didn't put them in the original ast16 PR because I didn't want to prolong it any further.

This is the first one. I've left a comment in the config file to reach out if there's a problem with the user "asterisk". I'm thinking for chan-dongle and chan-lantiq the permissions may probably not enough. But we could add the user to more groups via the init script (/lib/functions.sh has group_add_user() etc.). So if we get some feedback about what's needed it should be easy to add.

Kind regards,
Seb


micmac1:
removed "echo |" from before awk call as it's not needed.

dhewg:
Nice, I had that on my TODO as well ;)

procd has `procd_add_reload_interface_trigger` which can take care of the interface part. See
https://github.com/openwrt/packages/commit/b53fcc98b90a09235bfface928be81ea218d4a11
https://openwrt.org/docs/guide-developer/procd-init-scripts?s[]=procd&s[]=add&s[]=reload&s[]=interface&s[]=trigger
https://web.archive.org/web/20180411225312/http://wiki.prplfoundation.org/wiki/Procd_reference

I'm not sure if a similar solution exists for block-mount/ntq, but if that'd be possible there wouldn't be a need for a hotplug script at all.

And while at it, you can add a config reload trigger, so `luci-reload` restarts asterisk if its config file changed. See
https://github.com/openwrt/packages/commit/c5eb9e9e42433ddd437a5693ae5a7f2dc17cf41a

micmac1:
> Nice, I had that on my TODO as well ;)
> 
> procd has `procd_add_reload_interface_trigger` which can take care of the interface part. See
> [openwrt/packages@b53fcc9](https://github.com/openwrt/packages/commit/b53fcc98b90a09235bfface928be81ea218d4a11)
> [https://openwrt.org/docs/guide-developer/procd-init-scripts?s[]=procd&s[]=add&s[]=reload&s[]=interface&s[]=trigger](https://openwrt.org/docs/guide-developer/procd-init-scripts?s%5B%5D=procd&s%5B%5D=add&s%5B%5D=reload&s%5B%5D=interface&s%5B%5D=trigger)
> https://web.archive.org/web/20180411225312/http://wiki.prplfoundation.org/wiki/Procd_reference

Hello @dhewg 

Thanks for the feedback. We could add procd_add_reload_interface_trigger to the init script. Something like this in config: hotplug interfaces "int1:int2:int3" and then iterate over the interfaces and set up the triggers.

I don't think there's anything available for waiting for a mount point or ntpd. Those are things I like to check for before the PBX is started. Obviously I want the CDR time to be accurate. And I depend on external storage, so I want to make sure it's mounted.

So in my opinion it's either using procd triggers for hotplug and not adding mount point/ntpd stuff. Or using the hotplug script (I could extend it to allow specifying multiple interfaces separated by ":" for instance).

Let me know what you think.

> I'm not sure if a similar solution exists for block-mount/ntq, but if that'd be possible there wouldn't be a need for a hotplug script at all.
> 
> And while at it, you can add a config reload trigger, so `luci-reload` restarts asterisk if its config file changed. See
> [openwrt/packages@c5eb9e9](https://github.com/openwrt/packages/commit/c5eb9e9e42433ddd437a5693ae5a7f2dc17cf41a)

Well, this one I don't really like. I mean, personally I wouldn't want a PBX to restart just because I changed the config. I mean at some point yes, but not immediately. I'd rather that I can change the config and then reload the PBX at a time of my choosing. The point it, this is not some dns proxy where it doesn't hurt to reload right away. I don't want procd to interrupt running calls just because I changed the config. That's my opinion on this matter.

Kind regards,
Seb

micmac1:
Hi again,

I played around with procd interface triggers. I think it's a nice feature, but in my opinion it's not a hotplug replacement. A hotplug script starts something when a certain event occurs. For instance when you plug your Wifi USB dongle into your computer and that results in your OS configuring the interface properly and bringing it up. That's hotplug.

The procd interface triggers are only configured once a service is started. So some command gets executed (maybe "/etc/init.d/asterisk restart") when a certain interface condition occurs (maybe "interface.wan.up"). So the user cannot disable the asterisk autostart ("/etc/init.d/asterisk disable") and hope for a procd interface trigger to bring up the asterisk service, because the triggers are only set up when the service is started. With a hotplug script this is possible.

For now I have removed the hotplug script from this pull request. So this now only updates the init script and adds uci config (for init). Maybe we keep the hotplug/trigger stuff in the back of our heads and come back to that later.

I have added two branches to my repo if anybody is interested in poking around. They both add to this pull request.

[hotplug script](https://github.com/micmac1/telephony/tree/ast16-init-hotplug)
[procd interface triggers](https://github.com/micmac1/telephony/tree/ast16-init-triggers-II)

P.S.: The init script is now also checking if user/group exists.

micmac1:
Added a note about uci config to the commit message.

micmac1:
Added "--" to the logger variable to make it foolproof.

micmac1:
Hi all,

Stefan reminded me that /etc/asterisk/lantiq.conf needs to be readable. This is valid for the asterisk configuration in general, of course. So I added the line "chown -R "$user":"$group" /etc/$NAME" to the init script. And some lines to add the user "asterisk" to the "dialout" group (see below for the why).

He said that additionally

```
chmod 644 /lib/firmware/voice_ar9_firmware.bin
chown asterisk:asterisk /dev/vmmc*
```

would be required and that by default /dev/vmmc* belongs to root:root. So I checked around a bit. The firmware permissions need to be corrected because the kmod-ltq-vmmc package is currently simply copying the binaries, and this particular binary (voice_ar9_firmware.bin) comes with the permissions "-rw-------".  So I think if we changed the install line in ltq-vmmc  from

`$(CP) $(PKG_BUILD_DIR)/firmware/$(FW_SOURCE) $(1)/$(FW_DIR)/$(FW_TARGET)`

to

`$(INSTALL_DATA) $(PKG_BUILD_DIR)/firmware/$(FW_SOURCE) $(1)/$(FW_DIR)/$(FW_TARGET)`

it would work without us having the provide a workaround in the init script.

And then I saw that ltq-vmmc creates the device nodes via its init script. If we added a line to it to change the group to dialout

`group_exists dialout && chown :dialout /dev/vmmc*`

and made sure that the user "asterisk" is also part of this group (the new init script does that) then this should work out fine. The "dialout" group is part of the default OpenWrt groups file.

I would be interested in hearing the feedback from ltq-lantiq maintainer @blogic. Would you mind if we a changed group ownership of /dev/vmmc* to dialout and installed the firmware with "$(INSTALL_DATA)"?

As a side note I found [this forum post](https://forum.openwrt.org/t/solved-asterisk13-or-15-sqlite3-database-problem/36856/8) which shows that users do want to run Asterisk as a non-root service. So making their life easier by updating some of our stuff would be good.

Kind regards,
Seb

jslachta:
Thank you, Sebastian. With your contributions asterisk should be named better-asterisk-16.x.

micmac1:
Whoa, I didn't expect this to be merged anytime soon :-) And I would have been OK with that. On the other hand this is master, so we got a lot of time for feedback until release in 2020 ;-) Thanks Jiri!

jslachta:
It is time to move on. There is plenty of time to make asterisk even better, so I am considering this as a start.

PR #425, opened 2019-05-08
asterisk-chan-sccp: fix gcc8 compiles

micmac1:
Patch applied upstream. Fixes compiles for arc targets, which already
use new gcc8 toolchain.

Maintainer: @jslachta 
Compile tested: ath79, archs
Run tested: N/A (compile fix only)

Description:
Hello Jiri,

I checked for compile failures on the build bots. Only found sccp not compiling (on ARC, which uses new toolchain already). The rest of the packages (including chan-lantiq) compile fine.

Patch was applied upstream by Diederik.

Kind regards,
Seb


jslachta:
Thank you, Seb!

dkgroot:
This fix should hopefully no longer be needed (removed it from my recent PR #444 )

PR #430, opened 2019-06-17
dahdi-linux: bump to 3.0.0

### LINKED ISSUE #427: dahdi-linux: build fail in kernel 4.19
Maintainer: Vittorio Gambaletta <openwrt@vittgam.net>
Environment: armv7-a WRT3200ACM snapshot kernel-4.19

Description:
dahdi-linux will build fail in any kernel 4.19 compile. maybe upgrade to 3.0.0 will solve the problem?

micmac1:
This commit goes back from using a git checkout to using a regular
upstream tarball. Version bumped to 3.0.0 which works with kernels
equal to or greater than 4.15. Patches refreshed.

resolves #427

Maintainer: @jslachta 
Compile tested: ath79 against kernel 4.19.50
Run tested: don't have the hardware

Description:
Hi Jiri,

Petr sent a patch to openwrt-devel earlier today ([here](https://lists.infradead.org/pipermail/openwrt-devel/2019-June/017666.html)) that switches all but two targets to 4.19. Once this gets merged would be the right time to commit this bump PR.

I'll ping again once this happens.

Kind regards,
Seb

### DIFF
--- a/libs/dahdi-linux/Makefile
+++ b/libs/dahdi-linux/Makefile
@@ -9,19 +9,12 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=dahdi-linux
-PKG_VERSION:=2.11.1-20180111
-PKG_RELEASE:=2
+PKG_VERSION:=3.0.0
+PKG_RELEASE:=1
 
-#PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-#PKG_SOURCE_URL:=https://downloads.asterisk.org/pub/telephony/dahdi-linux/releases
-#PKG_HASH:=f59f382365118205e77d2874f1c0e1546e936247bcc45f07a43bc21778bee9df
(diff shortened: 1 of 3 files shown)


micmac1:
Hello Jiri,

The first targets have been changed today to default to kernel 4.19 (for instance ath79, mvebu & x86), so the dahdi-linux builds will now start to fail on the buildbots.

I've updated the PR by adding `@!(LINUX_4_9||LINUX_4_14)` to the kmod-dahdi depends. This will prevent any attempt to build kmod-dahdi (3.0.0) on the older kernels (4.9 and 4.14 are the only older kernels that upstream OpenWrt is using). In doing so we prevent build failures on the targets which haven't yet been updated to 4.19.

Please consider merging this PR.

Thanks!
Seb

EXSERENS:
Master branch start prepare for next release using kernel 4.19, now is good time to bump this to 3.0.0.
Maybe 3.1.0-rc1 :D

jslachta:
Good job. Thank you!

PR #445, opened 2019-07-31
yate: Rewrite init script to use procd

robho:

---

Maintainer: @jslachta
Compile tested: No
Run tested: Only on 18.06 branch


robho:
Yate hasn't been running properly for me. Seems like it sometimes dies. I'll check..

micmac1:
I had asked Robert to convert yate to procd. There is a patch on mailing list which makes procd default. But to be fair yesterday John Crispin replied saying he doesn't think it's a good idea to do that and that default should be sysv init scripts. So Robert thank you and no rush. We might as well wait this one out.

PR #452, opened 2019-08-31
[RFC] asterisk-17.x: add package

dangowrt:
Maintainer: @jslachta 
Compile tested: x86/legacy, x86/64, lantiq/xrx200 (incl. chan-lantiq)
Runtime tested: x86/legacy (incl. DAHDI with HFC-S PCI), x86/64

Description:
Add package for asterisk-17.0.0-rc2 to prepare for upcoming Asterisk 17 release.
Package new prometheus support.

### DIFF
--- a/net/asterisk-16.x-chan-lantiq/Makefile
+++ b/net/asterisk-16.x-chan-lantiq/Makefile
@@ -27,12 +27,12 @@ PKG_FLAGS:=nonshared
 include $(INCLUDE_DIR)/package.mk
 
 define Package/$(PKG_NAME)
-  SUBMENU:=Telephony Lantiq
+  SUBMENU:=Telephony
   SECTION:=net
   CATEGORY:=Network
   TITLE:=Lantiq channel driver
   URL:=https://github.com/kochstefan/asterisk_channel_lantiq
-  DEPENDS:=+asterisk16 +kmod-ltq-vmmc
+  DEPENDS:=asterisk16 +kmod-ltq-vmmc
 endef
 
 define Package/$(PKG_NAME)/description
--- a/net/asterisk-17.x-chan-lantiq/Makefile
+++ b/net/asterisk-17.x-chan-lantiq/Makefile
@@ -1,0 +1,71 @@
+#
+# Copyright (C) 2018 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=asterisk17-chan-lantiq
+PKG_VERSION:=20190803
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_SOURCE_URL:=https://github.com/kochstefan/asterisk_channel_lantiq.git
+PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
+PKG_SOURCE_VERSION:=1d940b38cde0348dfe129d2b764e6faee440c45b
+PKG_MIRROR_HASH:=ff838ff2a4c5353fadd73806e1513f59f224914582b6ba004165712268bc94e5
+PKG_SOURCE_PROTO:=git
+
+PKG_LICENSE:=GPL-2.0
+
+PKG_MAINTAINER:=Jiri Slachta <jiri@slachta.eu>
+
+PKG_FLAGS:=nonshared
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+  SUBMENU:=Telephony
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Lantiq channel driver
+  URL:=https://github.com/kochstefan/asterisk_channel_lantiq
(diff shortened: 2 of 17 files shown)


micmac1:
Hello @dangowrt 

Asterisk 13 was removed not too long ago. This was done, from my point of view, because maintaining two Asterisk versions in each branch was a lot of work. For every security patch usually both had to be updated, in _every_ branch.

There is not a lot of people actively working the telephony repo. There are a few, but they're not all working on Asterisk either. Some may prefer yate or siproxd.

I think that having only one Asterisk version is the way to go. And in my opinion that version should be an LTS release, because chances are much higher that they'll cover the life cycle of a branch. Asterisk 16 is an LTS release and is new. 17 is not (won't be) an LTS release.

Just my 2 cents :-)

Kind regards,
Seb

dangowrt:
@micmac1 I agree, having more than one version of asterisk shipped in a release (like we did for LEDE 17.01 coming with Asterisk 11 and 13, and OpenWrt 18.06 coming with Asterisk 13 and 15...) is a bad idea and ends up being a lot of work for each discovered vulnerability.
However, staying with the LTS release even in the development branch would be a mistake, because then we will have to make big jumps and nobody will have the time for testing...
Hence, I'd suggest to ship OpenWrt releases with a single LTS version of Asterisk while following upstream as close as possible in our development branch.

micmac1:
@dangowrt 
I understand your argument, but as long as there is nobody stepping forward and saying "I would like to do the actual work" (keeping the package up to date) I don't see that the argument is relevant.

Also, I don't think we had big problems going from one release to another. Going from Asterisk 13 to 16 we did not have to fix multiple issues, there was just a _single_ blocker, namely adapting the module loader to musl. So it wasn't "a big jump" for us. Hence from my personal point of view I don't think "riding the LTS wave" is a mistake (like you phrased it).

Kind regards,
Seb

jslachta:
I am not entirely sure about this. The maintenance costs are quite high when we support two or more versions in released branches. From my point of view, we can actively support only one asterisk package for released OpenWrt. 

Overall I think you are both right. The existence of the latest asterisk package makes sense for the master branch as a transition package, which could help with the movement to the next LTS release.

I would vote to have two asterisk packages only in the master branch with the active support for the LTS package. When the latest nonLTS package breaks, we can easily mark it as @BROKEN. In the case of releasing OpenWrt the nonLTS package will be removed from the release branch.

micmac1:
Hi all, I'm not against adding a second release, but I wouldn't put any effort into PRs for it myself.

BTW, if this PR would get merged as-is, chan-lantiq wouldn't be built by the build bots anymore. Explanation can be found in its commit history.

jslachta:
Hello Daniel (@dangowrt),

from the user perspective, there was only one big jump, namely the time when new SIP channel was introduced. The remaining new modules does not give any key functionality that we would desperately need on OpenWrt. From my POV it is only nice to have.

From the maintainer perspective, the pjsip introduction was the first big jump. It has settled down with asterisk-13. I do not recall any other big jumps from asterisk-11 to asterisk-16. The most problematic part was fixing security issues for multiple asterisk versions within a few OpenWrt releases. Since the LTS package is the most reliable, stable and usable in production environment, I prefer having only this one in telephony repository. Having an extended upstream support is a key feature for us. It reduces the maintenance cost.

The only situation when we will have two separate versions of asterisk is in a transition from one LTS package to another LTS package - for a limited amount of time.

I am sorry for your work, but I am saying no to your pull request.

I hope you understand that.

Kind regards,
Jiri

dangowrt:
Now that asterisk-17 has officially been released, we can drop asterisk-13.x and keep only asterisk-16.x as LTS release and asterisk-17.x as bleeding-edge?

dangowrt:
So you are saying that you only want to have asterisk-16 even in master branch and replacing it with the most recent release (17.0.0) is not an option (we would of course keep asterisk-16.x in 19.07 release branch).
In that case you would definitely loose me doing testing and maintenance as in our group we were desperately waiting for Prometheus support arriving in asterisk and we have switched to 17.0.0 for that reason already. I don't mind continuing with local builds, but up to now at least I've been able to contribute back to OpenWrt whatever we found.

jslachta:
Hi, Daniel (@dangowrt),

I know the community awaits the Prometheus module. However, the advantages of having the LTS release of Asterisk really outweighs having the cutting edge, sometimes unstable Asterisk with the newest modules. The new functionality in cutting edge releases seemed to me as a customer technology preview. I am also awaiting this module, but I would prefer to wait for the new LTS release, where all the new functionalities will settle down and tune to perfection. 

I have seen several cases where users were installing the unstable, latest Asterisk instead of installing the stable one. When problems occurred, users were not happy and I was not happy either. Fixing multiple versions of one product, often in various branches - it multiplied the resulting work. Since then, I prefer having a rule one version of the package for one release (I also consider the master branch as a "kind of" release). It makes the user experience better. We should focus on enhancing the integration with OpenWrt. Take a look at the awaited HEP (HOMER) module - introduced in Asterisk 12, and this module was present in Asterisk 13 in OpenWrt. I think it is not that much later.

Moreover, we do not have the workforce to maintain that many packages.

I am sorry, but it is still no. The Prometheus module will make it to OpenWrt, but not this time - in the next LTS. 

Kind regards,
Jiri

PR #457, opened 2019-09-05
[18.06] asterisk-15.x: add patch for AST-2019-004

micmac1:
Add patch for a remote crash vulnerability. Crash can occur when
negotiating for T.38 with a declined stream.

CVE-2019-15297

Refreshed some other patches.

Maintainer: @jslachta 
Compile tested: 18.06 ar71xx
Run tested: master

Description:
Same CVE fix for 18.06. Had to fiddle with one patch hunk where the next line in our 15.x version returns a different result than in the patch. Should be fine. Refreshed some other patches.


micmac1:
Oh, this is it for AST-2019-04. We have no other vulnerable versions.

PR #466, opened 2019-10-05
libosip2: bump to 5.1.0

micmac1:
Version bump.

- add ABI_VERSION as per OpenWrt package policies [1]
- don't package version-less symlink also as per [1]
- add --disable-silent-rules to keep build verbose
- remove leftover whitespace
- drop patch now included in tarball

[1] https://openwrt.org/docs/guide-developer/package-policies

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: ath79 master dlink router

Description:
Hi Jiri,

Added ABI_VERSION. Depending pks will pick it up and depend on libosip2-12, where 12 is the SONAME. I tried this and it worked fine.

ABI_VERSION seems like a neat feature of OpenWrt. I'll add it in the future when I update libs.

Kind regards,
Seb


micmac1:
Oh, forgot to mention, run-tested this with siproxd. That's the only user of this lib AFAIK.

PR #476, opened 2019-10-22
baresip: add VP8 and VP9 module

jslachta:
Maintainer: me / @jslachta 
Compile tested: x86, master
Run tested: x86, master

Description:
This PR adds VP8 and VP9 module.


micmac1:
Simply "+libvpx" doesn't work? Gives a reverse dependency error? If that's the case maybe install both in one package and call it mod-vpx? Upstream seems to treat it as one judging by them using the same variable. Just saying .. :)

jslachta:
"+libvpx" works, it is a copy paste mistake.

PR #479, opened 2019-10-24
pjproject: sync patches with Asterisk 16.6.1

micmac1:
Quote from Asterisk commit e6cc1f5083bbb85a76280e4bbada7be265b0a106:

"Issues in pjproject 2.9 caused us to revert some of their changes as a
work around. This introduced another issue where pjproject wouldn't
build with older gcc versions such as that found on CentOS 6. This
commit replaces the reverts with the official fixes for the original
issues and allows pjproject to be built on CentOS 6 again."

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: ath79 master, dir-825-c1, asterisk using pjsip behind kamailio

Description:
As said earlier, here the PR for patch sync.


micmac1:
Updated commit message subject, removed leading "libs/".

PR #480, opened 2019-10-24
kamailio-5.x: add new modules + cleanup

micmac1:
From upstream release notes:

Six new modules:

 - app_lua_sr – old Lua API before introduction of KEMI
 - lost – HELD (RFC6155) and LOST (RFC5222) location-based routing
 - kemix – KEMI specific extensions
 - rtp_media_server – embedded RTP and media processing
   functionalities for Kamailio (like playing media or bridging in a
   B2BUA manner, includes support for different codecs, including Opus)
 - secfilter – filters to allow/block using whitelists/blacklist based
   on user agents, IP addresses, countries, domains and users
 - xhttp_prom – generates suitable metrics for a Prometheus monitoring
   platform, answering to Prometheus pull requests

Additionally this removes libxml2 dependency from modules, because base
kamailio package already depends on it.

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: ath79 master, dlink dir-825-c1, asterisk behind kamailio.

Description:
Hi Jiri,

Here some new modules plus removal of gratuitous libxml2 deps (well, deps are needed, but we don't need to specify them twice). I runtested my normal setup, not the new modules. One module I added as BROKEN because we don't have the deps available. If there's interest/need we can add later.

Regards,
Seb

### DIFF
--- a/net/kamailio-5.x/Makefile
+++ b/net/kamailio-5.x/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=kamailio5
 PKG_VERSION:=5.3.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://www.kamailio.org/pub/kamailio/$(PKG_VERSION)/src
 PKG_SOURCE:=kamailio-$(PKG_VERSION)$(PKG_VARIANT)_src.tar.gz
@@ -31,6 +31,7 @@ MODULES_AVAILABLE:= \
 	alias_db \
 	app_jsdt \
 	app_lua \
+	app_lua_sr \
 	app_python \
 	app_python3 \
 	app_ruby \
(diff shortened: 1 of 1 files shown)


micmac1:
I wrote "some new modules", but this is really _all_ new modules. ;-)

jslachta:
Thank you!

micmac1:
Thanks!

hnyman:
Where are the absent packages that are referenced after this was committed ????

```
perus@ub1910:/Openwrt/r7800$ make clean
WARNING: Makefile 'package/feeds/telephony/kamailio-5.x/Makefile' has a dependency on 'bcunit', which does not exist
WARNING: Makefile 'package/feeds/telephony/kamailio-5.x/Makefile' has a dependency on 'mediastreamer2', which does not exist
WARNING: Makefile 'package/feeds/telephony/kamailio-5.x/Makefile' has a dependency on 'ortp', which does not exist
 make[1] clean
```

Having the culprit submodule defined as BROKEN does not help for these config errors, I think.
Please remove / comment out the line that references non-existent packages.

PR #482, opened 2019-10-28
asterisk-16.x: remove libxml2 hacks

micmac1:
There were problems in the past with OpenWrt's libxml2 package (also
it's host package). This led to quite a few workarounds in the asterisk
Makefile.

Recent changes in the libxml2 package have cleared this up.

1. a proper xml2-config file is now installed by both libxml2 and
   libxml2/host
2. xml2-config is now installed with a host triplet prefix (e.g.
   "mips-openwrt-linux-xml2-config" for libxml2 and
   "x86_64-pc-linux-gnu-xml2-config" for libxml2/host), additionally
   these tools get a symbolic link to xml2-config

That means that now OpenWrt's libxml2 can be properly detected and the
workarounds are no longer required.

This commit also removes an explicit dependency on libxml2 from
res-calendar-caldav. The base package already depends on libxml2, hence
the dependency is implicit.

Maintainer: @jslachta 
Compile tested: ath79, master
Run tested: N/A, no runtime change
Description:
Hi Jiri,

Nothing major, just cleaning up. It's nice to get rid of some hacks, though :)

Regards,
Seb

### DIFF
--- a/net/asterisk-16.x/Makefile
+++ b/net/asterisk-16.x/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 AST_MAJOR_VERSION:=16
 PKG_NAME:=asterisk$(AST_MAJOR_VERSION)
 PKG_VERSION:=$(AST_MAJOR_VERSION).6.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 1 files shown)


jslachta:
We can get rid of those hacks, finally! Thank you! 👍

PR #503, opened 2020-01-11
asterisk-16.x: fix compile with PKG_ASLR_PIE

### LINKED ISSUE #502: [19.07, master] Asterisk 16.x is not build with ASLR PIE
Maintainer: @jslachta 

When PKG_ASLR_PIE is enabled for the build Asterisk 16.x can not be compiled due to this error. Most probably similar to [#9845](https://github.com/openwrt/packages/pull/9845) from packages feed.

Build log is same for OpenWrt 19.07 and OpenWrt master:
```
/-----/build/staging_dir/toolchain-arm_cortex-a9+vfpv3_gcc-7.5.0_musl_eabi/lib/gcc/arm-openwrt-linux-muslgnueabi/7.5.0/../../../../arm-openwrt-linux-muslgnueabi/bin/ld: ast_expr2fz.o: relocation R_ARM_MOVW_ABS_NC against `stdin' can not be used when making a shared object; recompile with -fPIC
ast_expr2fz.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile:182: recipe for target 'check_expr2' failed
make[4]: *** [check_expr2] Error 1
```
By default PKG_ASLR_PIE is in the default settings disabled and it is possible that it can be enabled in the future. 

As workaround till the issue is find and fixed, `PKG_ASLR_PIE:=0` should be applied to the Makefile.

micmac1:
CFLAGS aren't used when compiling objects for check_expr2. This commits
adds the flags, which fixes the compilation when PKG_ASLR_PIE is
enabled. Note: The STANDALONE define is removed because it is already
defined in _ASTCFLAGS.

Fixes #502

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: ath79 19.07, copied over the check_expr2 executable and checked a file from the tests directory.

Description:
Fix ASLR PIE build fail

Hi Jiri, happy new year! :)

### DIFF
--- a/net/asterisk-16.x/Makefile
+++ b/net/asterisk-16.x/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 AST_MAJOR_VERSION:=16
 PKG_NAME:=asterisk$(AST_MAJOR_VERSION)
 PKG_VERSION:=$(AST_MAJOR_VERSION).6.1
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=asterisk-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.asterisk.org/pub/telephony/asterisk/releases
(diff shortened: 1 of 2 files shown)


jslachta:
Thank you, Seb. I made the same change locally, but a little bit uglier, so I chose yours. Happy new year to you too!

micmac1:
Thanks! I'll send this upstream.

BKPepe:
Do you think that this can be included also in `OpenWrt 19.07`?

PR #509, opened 2020-03-19
dahdi-linux: bump to version 3.1.0

micmac1:
This commit provides compatibility with kernel 5.4.

Except for the version bump this commit updates the HFC-S PCI driver to
not rely on the macro "dahdi_pci_module" anymore, because it has been
removed upstream in commit 4af6f69.

Also, a patch is added that makes the include of "<linux/pci-aspm.h>"
conditional on the kernel version, because starting with kernel 5.4 this
header has been removed and its contents merged into "pci.h".

Maintainer: @jslachta 
Compile tested: x86_64
Run tested: N/A, don't have the hardware for it

Description:
Hi Jiri,

Some targets were upgraded to kernel 5.4. For these targets most of telephony feed is currently failing, because dahdi-linux won't compile ([link](https://downloads.openwrt.org/snapshots/faillogs/x86_64/telephony/)).

I found that upgrading from 3.0.0 to 3.1.0 brings us a whole lot closer to being able to compile dahdi-linux against kernel 5.4. Two smaller issues I found that also prevent a successful compile, so I fixed both in this commit as well.

Like this I can compile all of the telephony feed fine in full master environment. Well, for kamailio the other pull request is still needed, because otherwise it still fails due to libpq pkg-config file issue.

All in all we should be good with kernel 5.4.

Thanks!
Seb

### DIFF
--- a/libs/dahdi-linux/Makefile
+++ b/libs/dahdi-linux/Makefile
@@ -9,12 +9,12 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=dahdi-linux
-PKG_VERSION:=3.0.0
+PKG_VERSION:=3.1.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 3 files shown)


dangowrt:
Thanks!

PR #519, opened 2020-04-10
kamailio-5.x: update build setup and add Python 3 module

micmac1:
Maintainer: @jslachta 
Compile tested: ath79 master + x86_64 master
Run tested: ath79 openwrt-19.07, tested with asterisk and one phone registering to it via kamailio, the other directly. All OK, calls fine. Also checked kamctl works (listing DB, removing entry, adding entry).

Description:
Hi Jiri,

While checking into Python3 I saw that Alpine sets up the build a bit differently. I like it a lot more the way they do it. I know they're not cross-compiling like us, but replicating their setup works fine for us. Like this we can also remove the manual "make" calls, which is nice.

With regards to the Python 3 module. It builds fine, but I wasn't able to run-test it because I ran out of space when trying to install python3-light :D But I'm pretty sure it's fine.

There are some additional details in the commit messages.

Kind regards,
Seb

### DIFF
--- a/net/kamailio-5.x/Makefile
+++ b/net/kamailio-5.x/Makefile
@@ -21,8 +21,9 @@ PKG_LICENSE:=GPL-2.0+
 PKG_LICENSE_FILES:=COPYING
 PKG_MAINTAINER:=Jiri Slachta <jiri@slachta.eu>
 
-# Defining PKG_BUILD_PARALLEL to 1 would be a noop due to the way we call make
-#PKG_BUILD_PARALLEL:=1
+PKG_INSTALL:=1
+
+PKG_BUILD_PARALLEL:=1
 
 MODULES_AVAILABLE:= \
 	acc \
@@ -388,9 +389,6 @@ endef
 #   "ims_diameter_server" -> libkamailio_ims
 #   "carrierroute" -> libtrie
 
-# "lib_target" is specified in order for the modules to get a proper
-# RPATH, as otherwise they would not find the internal libraries.
-
 EXTRA_MODULES:= \
 	$(if $(CONFIG_PACKAGE_kamailio5-mod-carrierroute),,carrierroute) \
 	$(if $(CONFIG_PACKAGE_kamailio5-mod-ims-diameter-server),,ims_diameter_server) \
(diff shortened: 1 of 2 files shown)


micmac1:
Hi all,

Added usb stick for more space, installed python3-base/light there and setup symlinks to /usr/bin and /usr/lib. Works.

```
Fri Apr 10 18:44:46 2020 daemon.err kamailio[15575]:  0(15575) ERROR: app_python3 [python_support.c:156]: python_handle_exception(): mod_init: Unhandled exception in the Python code:
Fri Apr 10 18:44:46 2020 daemon.err kamailio[15575]: TypeError: Function 'my_mod_init' of module 'myscript' has returned None. Should be a class instance.
Fri Apr 10 18:44:46 2020 daemon.err kamailio[15575]:  0(15575) ERROR: app_python3 [app_python3_mod.c:423]: apy_load_script(): Error calling mod_init
Fri Apr 10 18:44:46 2020 daemon.err kamailio[15575]:  0(15575) ERROR: app_python3 [app_python3_mod.c:163]: mod_init(): failed to load python script
Fri Apr 10 18:44:46 2020 daemon.err kamailio[15575]:  0(15575) ERROR: <core> [core/sr_module.c:849]: init_mod(): Error while initializing module app_python3 (/usr/lib/kamailio/modules/app_python3.so)
```

I know you see an error :) But you need to setup a python script that kamailio can use and I am not bored enough to try to figure out how to do that.

I also tried the plugin when it was compiled with python-config without "--embed":

```
Fri Apr 10 18:52:13 2020 daemon.err kamailio[15622]:  0(15622) ERROR: <core> [core/sr_module.c:512]: load_module(): could not open module </usr/lib/kamailio/modules/app_python3.so>: Error relocating /usr/lib/kamailio/modules/app_python3.so: PyTuple_New: symbol not found
Fri Apr 10 18:52:13 2020 daemon.err kamailio[15622]:  0(15622) CRITICAL: <core> [core/cfg.y:3543]: yyerror_at(): parse error in config file /etc/kamailio/kamailio.cfg, line 124, column 12-27: failed to load module
```

So that bombed. The module really needs access to the lib.

Actually, it turned out that python-config doesn't add "-lpythonx.y" since Python 3.8 (what we have in master now). Now you're really supposed to add "--embed" if you require the linking (like here). So the patch looks good to me.

Kind regards,
Seb

micmac1:
Hi Jiri,

Updated the app_python3 patch with the one that just got accepted upstream. It provides backwards compat with Python < 3.8. But for our intends and purposes there is no change, because we have 3.8 already in master.

Best regards,
Seb

micmac1:
Force-pushed a change in the python3 patch (added a "2>/dev/null" like upstream did), to get CI to work. CI failed. I realized that I probably needed to rebase my commits on master to get CI configuration, so I did that. Let's see if it works :)

jslachta:
Seb, would you be so kind to rebase the branch on master?

micmac1:
>Seb, would you be so kind to rebase the branch on master?

Hi, I did, ci still failed. The packages repo isn't checked out currently, so we don't get our deps :/

jslachta:
The PR fixing CI merged. :-)

micmac1:
Cool, let me rebase. Two thumbs pressed :)

jslachta:
Hi Seb,

I like all the changes you have made. Thank you.

Issue #515, opened 2020-04-04
asterisk16-chan-iax2: WARNING: iax2/firmware.c:235 iax_firmware_reload: Error opening firmware directory '/usr/share/asterisk/firmware/iax': No such file or directory


532910:
Maintainer: @jslachta
Environment: 19.07.2 / sunxi / cortexa53 / sun50i-h5-nanopi-neo2
Description:
```
CLI> module load chan_iax2.so
...
WARNING: iax2/firmware.c:235 iax_firmware_reload: Error opening firmware directory '/usr/share/asterisk/firmware/iax': No such file or directory
```

An empty dir should be created to prevent this warning.
I don't use firmware or IAX provisioning.

PR #531, opened 2020-05-14
asterisk-16.x: bump to 16.10.0 + init/config changes

micmac1:
Maintainer: @jslachta 
Compile tested: master  ath79
Run tested: 19.07 ath79, made a few phone calls

Description:
Hi Jiri,

Besides the update I made a few changes to init and config. Don't know if you agree, so let me know. I can change/remove as needed. Commit message have my reasoning.

Closes: #515 

Kind regards,
Seb


micmac1:
That's weird. I want to check why circleci failed. So I click. But now it doesn't take me to the logs. Instead it wants me to login via github (oauth). And then github tells me that circleci want's read _and_ write access to my repos. That feels fishy. I'm not going to do that :/

micmac1:
Hello @BKPepe and @champtar 

Any idea why circleci doesn't let me view the log anymore, instead prompting me for oauth with write access? I guess it's like this for everybody now. Has anything expired?

Kind regards,
Seb

micmac1:
Found something is going wrong with the lua detection and fixed it. This was already a problem before 16.10.0, but we didn't see any build failures. It's only a problem in master, though.

jslachta:
Thank you, Seb. I am merging this.

PR #532, opened 2020-05-17
rtpengine: prevent picking up host system flags

micmac1:
lib/lib.Makefile looks for libsystemd and dpkg-buildflags. These are not
available in OpenWrt, but they could be picked up from the build host.

This commit extends 01-cflags.patch a bit to prevent that.

Maintainer: me
Compile tested: master ath79
Run tested: no runtime-change

Description:
Hi all,

Saw this going wrong on one build bot ("truecz-dock-01"). I'm betting it has dpkg-buildflags installed which introduces some host system flags into the build. The commit also prevents libsystemd detection, as a precautionary measure.

Kind regards,
Seb

### DIFF
--- a/net/rtpengine/Makefile
+++ b/net/rtpengine/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=rtpengine
 PKG_VERSION:=mr8.3.1.4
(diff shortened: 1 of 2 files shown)


micmac1:
This stuff doesn't really fit into the cflags patch. Changed it around a bit.

PR #534, opened 2020-05-23
freeswitch-stable: rename to freeswitch

micmac1:
Maintainer: me
Compile tested: ath79 master
Run tested: N/A (no changes except the name)

Description:
Hi all,

It's been a while since freeswitch was moved to abandoned. I hope it's OK to rename freeswitch-stable to freeswitch now? Please let me know if you think otherwise. I'll leave this open, to linger, for a while.

I've done the folder moves in separate commits, so that it's easier to see the changes done to the files. I didn't really edit manually, but mostly just used sed.

While renaming I also remove the use of $(PKG_NAME) from the Makefiles. I used it like this because for some reason I thought it was the way to go. But in hindsight I think this was a bit misguided and made the Makefiles less readable. So this pull request does away with that, too.


jslachta:
+1 from me. I was going to ask you some time before, but I totally forgot that.

micmac1:
>+1 from me. I was going to ask you some time before, but I totally forgot that. 

Thanks Jiri!

PR #537, opened 2020-06-09
asterisk-chan-dongle: bump + cleanup

### LINKED ISSUE #536: asterisk-chan-dongle "excess" dependencies
Maintainer: @micmac1 
Environment: (any)
Description:

The asterisk-chan-dongle package depends on:
```
+kmod-usb-acm +kmod-usb-serial +kmod-usb-serial-option +libusb-1.0 +usb-modeswitch
```

However, being and embedded environment, I consider these dependencies should not be included, as option devices don't need the usb-acm module (and vice versa?), and even then, maybe the user included the driver in kernel? So installing the correct serial driver should be left to the user, the same goes with the other dependencies (I don't know if usb-modeswitch is needed for all modems).

Thank you for your great work

micmac1:
Bump dongle plugin to latest git revision. PKG_SOURCE_SUBDIR and
PKG_BUILD_DIR are removed because defaults are fine. PKG_VERSION is
replaced by PKG_SOURCE_DATE, which is in line with OpenWrt
recommendation.

Soft deps are removed as per request in issue #536.

iconv handling is updated. Upstream now lets us specify the include
directory, so that makes the patch simpler. The iconv lib detection is
also updated so that it never picks the libc iconv. This way we get the
same iconv handling independent of the libc in use. Now it'll be always
iconv-stub or iconv-full.

libsqlite3 dep is made explicit. There was always a dep on it, but it
was implicit via asterisk. When the dep is made explicit, the dongle
package picks up on the sqlite3 ABI_VERSION, which is nice.

Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: no, don't have the hardware

Description:
Version bump and cleanup. This would also resolve #536, as the soft deps are removed.

Resolves #536 

Kind regards,
Seb

### DIFF
--- a/net/asterisk-chan-dongle/Makefile
+++ b/net/asterisk-chan-dongle/Makefile
@@ -8,23 +8,19 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=asterisk-chan-dongle
-PKG_VERSION:=1.1-20180619
-PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=https://github.com/wdoekes/asterisk-chan-dongle.git
-PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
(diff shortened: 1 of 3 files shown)


jslachta:
Thank you, Sebastian. Perfect!

PR #540, opened 2020-06-11
asterisk-chan-dongle: fix iconv lib detection

micmac1:
A small but important oversight crept into the last asterisk-chan-dongle
update. The iconv library detection was changed to look for iconv()
instead of libiconv().

Reminder: the autoconf macros for finding libraries do _not_ include any
extra headers when testing the compile. So the redefinitions done in
iconv.h provided by any libiconv (stub or regular) aren't visible during
the test the macros execute.

This means that when testing for iconv() the only iconv that can be
detected is the one from libc. Moreover, if the libc doesn't provide
iconv, the check will fail. Below example is with uclibc:

checking for iconv in -liconv... no
configure: error: iconv library missing
Makefile:87: recipe for target '/builder/shared-workdir/build/sdk/build_dir/target-arc_archs_uClibc/asterisk-chan-dongle-asterisk16/asterisk-chan-dongle-2020-05-28-328b2b7d/.configured_a17fb5ef857664f03cd0ce37cc5ea591' failed

So, this changes the patch to look for libiconv(), again.

Also, instead of using AC_CHECK_LIB this uses AC_SEARCH_LIBS, like
upstream does. There apparently was a problem with that in the past, but
now it works fine.

checking for rm... rm
checking for library containing libiconv... -liconv
checking for sqlite3_open in -lsqlite3... yes

Maintainer: @jslachta 
Compile tested: ath79 master with nls on and off, arc38s (uclibc) SDK
Run tested: no such hardware

Description:

Hi Jiri,

I messed up iconv in previous commit, at least for uclibc. Here's the fix :)

Kind regards,
Seb


jslachta:
Thank you!

PR #555, opened 2020-08-05
libks: fix compilation without deprecated OpenSSL APIs

neheb:

Maintainer: @micmac1 
Compile tested: ath79

### DIFF
--- a/libs/libks/Makefile
+++ b/libs/libks/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=libks
 
 PKG_VERSION:=1.6.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


micmac1:
Hi Rosen,

This should go upstream first, to get a proper review and (hopefully) get commited, so that it then can trickle down to the downstream.

I'm not sure what your expectation here is. Do you expect me to blindly commit this or do you expect that I take it upstream?

Kind regards,
Seb

neheb:
The former.

micmac1:
I'd have to decline. I think it's too risky to take these changes lightly. They should be properly reviewed by somebody who understands the patches. This should be done upstream.

neheb:
Looking at this patchset, it's just missing headers, getm* API rename, and removal of explicit initialization/deinitialization. The last one is completely deprecated as it is already done internally.

I submitted it anyway: https://github.com/signalwire/libks/pull/81

neheb:
This was merged upstream.

micmac1:
Thanks!

PR #559, opened 2020-08-07
freeswitch: bump to 1.10.4

micmac1:
Maintainer: me
Compile tested: ath79 master
Run tested: not yet, will do before commiting

Description:
Update to FS 1.10.4. It's a bit more involved than usual. Upstream split out sofia-sip library into separate package. So this gets added here. They did the same with spandsp. But their spandsp "fork" may not be a drop-in replacement for the original spandsp. So I reverted some of the changes to keep spandsp contained within freeswitch, like it was before. This we can revisit in the future.

I'll runtest this when I get some time to do it, hopefully in the next couple of days.

Kind regards,
Seb


micmac1:
Run-tested this today. Made a few calls. Also originated a fax from FS to a handheld. Obviously fax could not be sent, but I heard the fax tone from FS, so that seems ok.

PR #566, opened 2020-09-04
asterisk-16.x: bump to 16.13.0 

dhewg:
Maintainer: @jslachta 
Compile tested: omap
Run tested: omap, calling and receiving calls work as expected

This bumps pjproject for asterisk, then asterisk itself.


micmac1:
Hi @dhewg 

Nice, thanks!

Small nitpick: I understand pjsip upstream now points their download to github. That's OK, but in our Makefile I probably wouldn't use https://github.com/pjsip/$(PKG_NAME)/archive. Instead I would likely go for https://codeload.github.com/pjsip/pjproject/tar.gz/$(PKG_VERSION)?.

Looking at the Makefiles in the packages repo that fetch the archive they need from github, most (all?) seem to use codeload. I think the reason is that this is a stable download link in the sense that the checksum never changes. If on the other hand you use the archive link I would expect that the checksum may change over time.

Kind regards,
Seb

neheb:
@micmac1 not correct. The reason is that a URL like https://github.com/pjsip/pjproject/archive/2.10.tar.gz redirects to codeload anyway. AFAIK, there's an unofficial policy to avoid using redirects.

codeload is not stable at all. I remember a go project that relied on submodules randomly having the codeload tarball change. The actual difference was an extra character in the short hash.

micmac1:
>@micmac1 not correct. The reason is that a URL like https://github.com/pjsip/pjproject/archive/2.10.tar.gz redirects to codeload anyway. AFAIK, there's an unofficial policy to avoid using redirects.

Meh, OK, then it doesn't matter.

dhewg:
Ok, I'll leave it as-is then :)

jslachta:
@dhewg, thank you for your contribution!

I will do some runtesting before the merge, looks good for the first look.

jslachta (approved):

jslachta:
Thank you!

PR #567, opened 2020-09-05
siproxd: update to version 0.8.3, improve logging and memory usage

guidosarducci:
**Maintainer**: @jslachta 
**Compile tested**: malta/mips_24kc/master & ea6350v3/ipq40xx/19.07.3
**Run tested**: malta/mips_24kc/master & ea6350v3/ipq40xx/19.07.3

**Description**:
This release includes bug fixes and improvements accumulated over the last
4 years of development snapshots [1].

Bump upstream version, build new plugins 'stats' and 'blacklist', and add
an 'sqlite3' package dependency for siproxd-mod-blacklist.

Disable procd capturing stderr to syslog, since siproxd in 'foreground'
mode writes both to syslog and stderr, duplicating log messages.

Include a patch to drastically reduce memory usage (RSS from 17MB to 5MB
on ipq40xx platform):

  * 005-reduce-rtpproxy-urlmap-size.patch

Drop the following patches now included upstream:

  * 010-syslog-msg.patch
  * 020-gcc10.patch
  * 100-musl-compat.patch

[1] http://siproxd.sourceforge.net/index.php?op=changelog.

Signed-off-by: Tony Ambardar <itugrok@yahoo.com>

PR #575, opened 2020-09-22
siproxd: add max clients configuration and default

guidosarducci:
**Maintainer**: @jslachta 
**Compile tested**: malta/mips_24kc/master
**Run tested**: malta/mips_24kc/master

**Description**:
Add a configuration option SIPROXD_MAX_CLIENTS used for compile-time sizing
of the maximum supported proxy clients. Set this to a default value of 32,
which is more suitable for a small network or CPE router and uses much
less memory than the upstream value of 512 (e.g. RSS of 2 MB vs 17 MB).

Drop the following patch in favour of the new config option:

  * 005-reduce-rtpproxy-urlmap-size.patch

Also fix an incorrect configure option, and silence many warnings due to
unused VCS 'ident' variables in the sources.

Signed-off-by: Tony Ambardar <itugrok@yahoo.com>


guidosarducci:
@micmac1 Hi Seb, this follows up from our recent discussion in https://github.com/openwrt/telephony/pull/567#issuecomment-687658932. Seems to work fairly well.

micmac1:
Hi Tony, I don't use siproxd myself so I'll gladly take your word for it :-) Have a good day and thanks for your work!

micmac1:
One small nitpick, maybe PKG_CONFIG_DEPENDS should be outside of the package definition? Maybe it works inside but usually it's set outside, no? I'll double-check tonight or tomorrow.

Am September 22, 2020 8:55:16 AM UTC schrieb guidosarducci <notifications@github.com>:
>@micmac1 Hi Seb, this follows up from our recent discussion in https://github.com/openwrt/telephony/pull/567#issuecomment-687658932. Seems to work fairly well.
>

guidosarducci:
Hmm. Yes, I think you're right: it does work but cleaner as you suggested. Also fixed some build warnings I noticed. Thanks!

micmac1:
Looks good to me, compiles fine. I'm wondering why spaces instead of
tabs in lines 2 and 3 of the $(ESED) command. That looks a bit weird to
me. But OK, I'm just nitpicking again :)

Thanks!

guidosarducci:
Good catch, I'm glad someone is paying attention! 🙄  Fixed up the whitespace...

micmac1:
Thanks a bunch!

jslachta (approved):

jslachta:
Thank you, guys!

PR #576, opened 2020-10-03
re+baresip: bump to 1.0.0

micmac1:
Maintainer: @jslachta 
Compile tested: master ath79
Run tested: run-tested on 19.07 ath79

Description:
Update to 1.0.0


micmac1:
Added two more commits that remove postinst scripts from baresip and restund. Now there's no need for clumsy chown calls. In baresip we can use fakeroot for that and in restund there's actually no need for it anyway.

Also updated first post to reflect run-testing was fine. The FILE_MODES feature isn't available in 19.07, so I verified the file modes by extracting the package as root and made sure that with the same file modes it runs well on 19.07.

Note: it may be that after adding the last commit CircleCI fails. I imagine there's a chance of that because upstream removed www.creytiv.com and restund still points PKG_SOURCE_URL there. As far as I can see upstream didn't fully migrate restund to github yet like they did with libre and baresip. So the download from within CircleCI of restund may fail, but it will work fine on OpenWrt build bots as the tar file is still archived there.

Kind regards,
Seb

micmac1:
Hi @alfredh :) Will restund's release also be migrated to github eventually?

Thanks!
Seb

jslachta (approved):

jslachta:
Thank you, Seb!

PR #583, opened 2020-10-09
dahdi-tools: remove glibc specific DECLS

neheb:
Fixes compilation with musl 1.2.0 and above.

Maintainer: @VittGam 
Compile tested: ath79

### DIFF
--- a/libs/dahdi-tools/Makefile
+++ b/libs/dahdi-tools/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dahdi-tools
 PKG_VERSION:=3.1.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


neheb:
ping @openwrt/telephone-maintainers

micmac1:
> ping @openwrt/telephone-maintainers

This looks like something that is not specific to OpenWrt and should go upstream. Have you already send it there?

neheb:
Upstream seems dead

micmac1:
> Upstream seems dead

Upstream is asterisk project. I doubt they're dead.

neheb:
Sure but last commit from 2018.

micmac1:
It doesn't mean they're dead. DAHDI is definitely not fast-paced, it's about legacy hardware IMHO. There's is a dahdi-linux/next branch with commits from this year. I don't think they accept PRs via github. Probably they would require that you submit patches via the same mechanism that one would use for asterisk itself ([Link](https://wiki.asterisk.org/wiki/display/AST/Patch+Contribution+Process)). It's a bit tedious to do (speaking from experience).

I could understand if you would not want to bother with this. In that case I'd look at this myself and post a patch upstream eventually. But I'd do it after musl 1.2 is commited to OpenWrt. I don't see any urgency here.

PR #587, opened 2020-11-06
[19.07] asterisk-16.x: fix AST-2020-001 and 002

micmac1:
Patches used:

http://downloads.asterisk.org/pub/security/AST-2020-001-16.diff
http://downloads.asterisk.org/pub/security/AST-2020-002-16.diff

Patch AST-2020-002-16.diff was amended a bit in res/res_pjsip_session.c:

if (++session->authentication_challenge_count > MAX_RX_CHALLENGES) {
	ast_debug(3, "%s: Initial INVITE reached maximum number of auth attempts.\n", ast_sip_session_get_name(session));
	return PJ_FALSE;
}

The above was not possible, because of missing bits introduced only in a
later version of Asterisk 16 (see upstream commit [1]). So the ast_debug
call was simplified accordingly.

Both patches were refreshed within OpenWrt SDK.

[1] https://github.com/asterisk/asterisk/commit/6abf6f345dbd0510d8a217d16cc1819e4d2bf815

Maintainer: @jslachta 
Compile tested: 19.07 SDK ath79
Run tested: 19.07 ath79, made a few calls. No problem seen.

Description:
Had to make a small amendment to upstream patch as mentioned in commit message. But shouldn't be a problem.

Kind regards,
Seb

### DIFF
--- a/net/asterisk-16.x/Makefile
+++ b/net/asterisk-16.x/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 AST_MAJOR_VERSION:=16
 PKG_NAME:=asterisk$(AST_MAJOR_VERSION)
 PKG_VERSION:=$(AST_MAJOR_VERSION).3.0
-PKG_RELEASE:=7
+PKG_RELEASE:=8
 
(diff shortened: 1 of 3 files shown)


micmac1:
Run-tested. First post updated.

PR #592, opened 2020-12-05
kamailio-5.x: build fixes

micmac1:
Maintainer: @jslachta 
Compile tested: archs38 + rtl838x
Run tested: don't have the hardware, but these are build-fixes only anyway.

Description:
Hi Jiri,

I was going around https://downloads.openwrt.org/snapshots/faillogs/ looking for builds that fail. I found Kamailio failing on uClibc targets due to Erlang (I expect the same for glibc). Kamailio also fails to build for the new rtl838x target. Fixes for both issues are included here.

I found one more build failure: asterisk-chan-sccp on uClibc targets. This is because OpenWrt doesn't build uClibc with XLOCALE support, I think. I can live with that :)

Kind regards,
Seb

### DIFF
--- a/net/kamailio-5.x/Makefile
+++ b/net/kamailio-5.x/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=kamailio5
 PKG_VERSION:=5.4.0
-PKG_RELEASE:=2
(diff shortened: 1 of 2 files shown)


jslachta (approved):

jslachta:
Looks good, thank you. 👍

PR #596, opened 2020-12-23
asterisk: micro version bump to 18.1.1

micmac1:
This addresses CVE-2020-35652 as described in:

https://downloads.asterisk.org/pub/security/AST-2020-003.html
https://downloads.asterisk.org/pub/security/AST-2020-004.html

Maintainer: @jslachta 
Compile tested: No, I just confirmed that our patches apply fine (they do). CI will confirm the build.
Run tested: No, this is a micro version update that only includes one added patch (which is tiny).

Description:
Hi Jiri,

They announced AST-2020-003 and AST-2020-004 today, for which they received CVE-2020-35652. This is related to History-Info header support, which they added in 2019 (correction: in September 2020, actually). I checked in 19.07 and there we do not have History-Info header support yet, so we don't need to worry about 19.07 and 18.06. The only package that needs attention is in master.

Kind regards and happy holidays to you, too!

Seb

### DIFF
--- a/net/asterisk/Makefile
+++ b/net/asterisk/Makefile
@@ -304,7 +304,6 @@ MODULES_AVAILABLE:= \
 	res-stun-monitor \
 	res-timing-dahdi \
 	res-timing-pthread \
-	res-timing-timerfd \
 	res-xmpp
 
 UTILS_AVAILABLE:= \
@@ -493,7 +492,7 @@ AST_CFG_FILES:= \
 AST_EMB_MODULES:=\
 	app_dial app_echo app_macro app_playback \
 	func_callerid func_logic func_strings func_timeout \
-	pbx_config res_crypto
+	pbx_config res_crypto res_timing_timerfd
 
 define Package/$(PKG_NAME)/install
 $(call Package/$(PKG_NAME)/install/lib,$(1),libasteriskssl)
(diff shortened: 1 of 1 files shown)


micmac1:
Hello @jslachta :-)

I'm thinking that maybe it would be a good idea to include the timerfd module in the base asterisk package. [1] mentions that the timing modules are used for iax2 and also (maybe) when playing files. Now we know that they're also used for DTMF, which is basic functionality that doesn't come with a module.

We have some modules that depend on timing modules. chan-iax2 depends on timerfd. res-fax on pthread. I checked AST_MODULE_INFO of these modules and there is no mention of any dependency on any timing modules. So it's not readily apparent what functionality depends on any of the timing modules.

Adding a timing module to the base asterisk package seems to make sense to me. Like app_dial for instance, which is also in the base asterisk package.

In [2] there is some more info from upstream on this topic. All in all we can say that we can't use kqueue, as it is meant for BSDs and Mac. pthread uses a userspace thread which is said to be not very efficient. timerfd looks like the best choice, it's timing comes from the kernel. All our kernels support it as far as I know. DAHDI doesn't seem like a good choice for obvious reasons.

The timerfd ipk weighs in at a little under 4KB (IPK). So it's not big. I could add a commit to the 18.1.1 pull request that adds it to the base package. What do you think?

Kind regards,
Seb

[1] https://wiki.asterisk.org/wiki/display/AST/Timing+Interfaces
[2] http://lists.digium.com/pipermail/asterisk-users/2015-February/285804.html

micmac1:
Sorry, wanted to make this comment in the issue raised yesterday. Will edit here and remove it and paste it over there :/

micmac1:
I compile and run-tested this in the meantime. Cheers!

jslachta:
Thank you so much!

PR #599, opened 2020-12-31
dahdi-tools: remove broken patch, disable unused features

### LINKED ISSUE #598: dahdi-tools does not compile under glibc
Maintainer: maybe @micmac1 
Environment: glibc

https://downloads.openwrt.org/snapshots/faillogs/arc_archs/telephony/dahdi-tools/compile.txt

micmac1:
"010-fix-non-glibc-builds.patch" is broken. This PR removes it and instead adds some configure args that avoids building some tools we don't package anyway.

Maintainer: @VittGam 
Compile tested: archs38
Run tested: no

Description:
Closes #598


micmac1:
Reviewed this some more. The patch can be removed altogether if some extra configure args are specified. This is less invasive so we should opt for this in my opinion.

neheb:
This seems to be a fine alternative to https://github.com/openwrt/telephony/pull/583

PR #601, opened 2021-01-01
dahdi-linux: fix compilation with musl 1.2.x

neheb:
Placing oslec.h in the normal include directory helps packages find it
properly. Also install to InstallDev.

Maintainer: @VittGam @micmac1 
Compile tested: powerpc


neheb:
weird. this is actually totally wrong.

Issue #600, opened 2021-01-01
freeswitch patch 230 needs refreshing


neheb:
Maintainer: @micmac1 
Environment: CentOS 7

There was some strange issue regarding that patch. make was failing on freeswitch. I used V=s to find out why and apparently it asked me whether or not I want to assume -R when patching.

V=s also showed that patch to have fuzz and wrong offsets.

PR #603, opened 2021-01-01
freeswitch: fix compilation with musl 1.2.0

neheb:
Fix patch 230 as it's a patch of a patch with fuzz and wrong offsets.

Refresh other patches.

Maintainer: @micmac1 
Compile tested: powerpc

### DIFF
--- a/net/freeswitch/Makefile
+++ b/net/freeswitch/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freeswitch
 PKG_VERSION:=1.10.5
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_MAINTAINER:=Sebastian Kemper <sebastian_ml@gmx.net>
 
 PKG_SOURCE:=freeswitch-$(PKG_VERSION).-release.tar.xz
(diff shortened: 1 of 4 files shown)


neheb:
Fixes: https://github.com/openwrt/telephony/issues/600

micmac1:
Hi there,

This commit removes the AC_TRY_RUN->AC_COMPILE_IFELSE replacement. And the subject line is wrong. Do you want to take another crack at this or do you want me to do it? Apart from that it would be nice to know how it actually fails for you (reproducer).

Kind regards,
Seb

micmac1:
Happy New Year by the way!

neheb:
> Hi there,
> 
> This commit removes the AC_TRY_RUN->AC_COMPILE_IFELSE replacement.

Upstream actually changed the makefile. It additionally checks whether or not it's being cross compiled making it no/op. Which is why I removed the section.

> And the subject line is wrong.

Meh I thought the issue was with musl 1.2.x. I don't actually know what the core issue is. My guess is some old version of some tool that comes with CentOS 7.

> Do you want to take another crack at this or do you want me to do it?

I'll edit the commit subject.

> Apart from that it would be nice to know how it actually fails for you (reproducer).
> 
> Kind regards,
> Seb

I think it's actually some issue with CentOS 7's quilt or something. It doesn't like fuzz.

micmac1:
Makes sense. Thanks!

PR #609, opened 2021-01-27
asterisk: disable uriparser

BKPepe:
Maintainer: @jslachta

Uriparser is a compliant URI parsing and handling library for RFC 3986.
For now, this is not included in OpenWrt feeds, but in Turris OS
uriparser is used as dependency for Updater-ng.

When Asterisk finds in build system there is uriparser or anything else mentioned in configure,
it tries to enable it by default. This applies to every package in
OpenWrt and because of that new packages which are added to OpenWrt tries to disable almost
everything by default. Because if someone adds library, some packages
are not compiled or tested.

Fixes:
Package asterisk is missing dependencies for the following libraries:
liburiparser.so.1

Compile tested: Turris Omnia, Turris OS 6.0 (unreleased) based on the latest OpenWrt master (feeds, which were used [here](https://repo.turris.cz/hbd/omnia/packages/git-hash)) and liburiparser [here](https://gitlab.nic.cz/turris/turris-os-packages/-/blob/develop/libs/uriparser/Makefile). It can be reproduced that liburiparser is added to the build system, compiled first and then try to compile asterisk.

### DIFF
--- a/net/asterisk/Makefile
+++ b/net/asterisk/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=asterisk
 PKG_VERSION:=18.1.1
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=asterisk-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.asterisk.org/pub/telephony/asterisk/releases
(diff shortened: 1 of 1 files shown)


micmac1:
Hi Josef,

If you asked me the commit message is long, winding and unclear. Maybe something short and to the point would suffice.

Apart from that the maintainer is @jslachta :-)

Kind regards,
Seb

BKPepe:
Hello Seb,

Ok! I will do it.

I like everything complicated and long.

Regards,
Josef

micmac1:
:-)

jslachta (approved):

jslachta:
Compile and run tested on mvebu/cortexa53.

Thank you, Josef.

BKPepe:
That was fast! I didn't have even chance to edit commit message, but never mind. Next time, it will be better.

PR #610, opened 2021-02-01
dahdi-linux: patch to compile with linux >= 5.6

lipnitsk:
Maintainer: @micmac1 
Compile tested: mt7621, Linux 5.10

Relevant kernel commits:
https://github.com/torvalds/linux/commit/4e24566a134ea167441a1ffa3d439a27cf400880
https://github.com/torvalds/linux/commit/d56c0d45f0e27f814e87a1676b6bdccccbc252e9

### DIFF
--- a/libs/dahdi-linux/Makefile
+++ b/libs/dahdi-linux/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=dahdi-linux
 PKG_VERSION:=3.1.0
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.asterisk.org/pub/telephony/dahdi-linux/releases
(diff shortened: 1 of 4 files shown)


micmac1:
Hello Ilya,

Thanks for the pull request. However, I'd be even happier if you could replace the patches with the related upstream commits (including headers).

Upstream: git://git.asterisk.org/dahdi/linux

The commits you need are in the "next" branch. The commits needed are these (in this order):

```
34b9c77c9ab2794d4e912461e4c1080c4b1f6184: Use proc_ops on kernels >= 5.6
4df746fe3ffd6678f36b16c9b0750fa552da92e4: Remove support for 32-bit userspace with 64-bit kernel
6d4c748e0470efac90e7dc4538ff3c5da51f0169: Remove checks for HAVE_UNLOCKED_IOCTL for kernel >= 5.9
```

If you have an issue doing this let me know.

Kind regards,
Seb

lipnitsk:
Thanks for the prompt response. It looks like the `next` branch has other useful warning fixes and may replace some other existing patches. What do you think about using tip of `next` and reducing the number of patches?

micmac1:
On Mon, Feb 01, 2021 at 05:21:51PM +0000, Ilya Lipnitskiy wrote:
> Thanks for the prompt response. It looks like the `next` branch has
> other useful warning fixes and may replace some other existing
> patches. What do you think about using tip of `next` and reducing the
> number of patches?

I'd rather upstream released a new stable version once they are happy
with what they have in next. We should stay away from dev branches and
stick to the releases IMHO.

micmac1:
Excellent, thanks for your work. Would you mind increasing the PKG_RELEASE in the Makefile to 4 so that we can merge this? I should've mentioned this earlier :-)

lipnitsk:
Oops - should have known better - sorry - still a bit new to this :)

micmac1:
> Oops - should have known better - sorry - still a bit new to this :)

You did great. Hope to hear from you again in the future!

Kind regards,
Seb

PR #621, opened 2021-03-07
pjproject: add fix for AST-2021-005 + brush up our Makefile

micmac1:
Maintainer: @jslachta 
Compile tested: ath79 master
Run tested: Not really, I only run-tested the AST-2021-005 on 19.07.

Description:
Hi Jiri,

The other AST-2021-* notices are covered in asterisk 18.2.2, so once that's upgraded we're done for master. Would be great if these then could be cherry-picked to openwrt-21.02.

Kind regards,
Seb


jslachta:
Thank you, Sebastian!

PR #648, opened 2021-06-04
yate: bump to 6.3.0 and enable speex codec

robho:
New yate modules in this release:
yate-mod-ilbcwebrtc
yate-mod-speexcodec
yate-mod-wiresniff

Some patches were removed:
110-no-libnsl.patch - not needed
120-create-thread-key-on-access.patch - not needed
150-video-codecs-mapping.patch - already applied upstream

Some additional changes:
* An upstream fix (r6470) is cherry-picked to avoid crashing when YATE
  receives an invalid SIP message.

* Build directory changed to include package version.

* Set PKG_BUILD_PARALLEL to allow parallelized build of yate.

* Speed up build if yate-mod-ilbccodec or yate-mod-sqlitedb is
  disabled (disable ilbc and sqlite in configure step).

* Removing umbrella package yate-collection-basic. It isn't built and
  hasn't been built for years and I think the modules it wants to
  install are a bit strange and probably not useful to most
  users. From the name yate-collection-basic I would expect to get
  some basic functionality, but why would I then need
  yate-mod-dumbchan, yate-mod-rmanager and yate-mod-tonegen? What is
  the use case? I think it's better that users themselves install the
  packages needed for their use case, especially in OpenWrt where
  flash space usually is limited and you don't want to install more
  packages than what you really need.

Maintainer: @jslachta
Compile tested: Netgear R6220, ramips, mipsel_24kc
Run tested: Not tested on a build from master, but I've tested yate 6.3.0 on 21.02.0-rc2


micmac1:
Hello @robho 

Please update the commit subject to use a lowercase letter after the colon (replace "Bump" with "bump"). Everything else looks good to me.

Thanks!
Seb

P.S.: powerpc_8540 test build fails, but there's a problem with this SDK, it doesn't build right now and the old one that is likely used instead has other issues.

robho:
Thanks for taking a look at the failing test. I'll leave it as it is then.

I've updated the commit's title as requested.
Have a nice weekend

jslachta (approved):

jslachta:
Hello, Robert (@robho),

thank you for taking a look at Sebastian's (@micmac1) notes.

I have tested your changes, and it looks good. I have no further comments.

Thank you so much!

Kind regards
Jiri

PR #649, opened 2021-06-07
yate: increase max acceptable size of incoming SIP messages

robho:
Some SIP UAs support lots of features and codecs which results in
large SIP messages. YATE, with its default configuration, truncates
and fails to parse received SIP messages which are larger than 1500
bytes. Let's increase the default max message size for OpenWrt users
to make it easier to use yate out-of-the-box.

The new max size of 8192 bytes has been arbitrarily chosen.

I've seen the SIP UA baresip produce messages larger than 1500 bytes
with its default configuration when authentication is used.

Maintainer: @jslachta 
Compile tested: No
Run tested: Only by manual configuration change. An increased maxpkt size works well for me. Verified with SIP UA baresip that had problems with YATE before the maxpkt increase.

### DIFF
--- a/net/yate/Makefile
+++ b/net/yate/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=yate
 PKG_VERSION:=6.3.0-1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


jslachta (approved):

jslachta:
Thank you, Robert (@robho)!

PR #657, opened 2021-07-15
kamailio: prevent arm -march overrides

micmac1:
src/Makefile.defs adds "-march=..." to target flags for ARM targets.
This can clash with the OpenWrt target flags in
CONFIG_TARGET_OPTIMIZATION. If for example the latter is set to
"-mcpu=cortex-a9" the following warning is issued:

cc1: warning: switch '-mcpu=cortex-a9' conflicts with '-march=armv5t' switch

This commit removes these flags. The addition is unwanted when
cross-compiling.

An issue was raised for this recently, see [1].

[1] https://github.com/openwrt/packages/issues/16135

Maintainer: @jslachta 
Compile tested: CI
Run tested: No

Description:
Hello @ByteEnable

Please try this and provide feedback. Thanks.

Regards,
Seb

### DIFF
--- a/net/kamailio/Makefile
+++ b/net/kamailio/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=kamailio
 PKG_VERSION:=5.5.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


neheb:
```
Applying ./patches/070-dont-override-arm-march.patch using plaintext:
patching file src/Makefile.defs
Hunk #1 FAILED at 1260.
Hunk #2 FAILED at 1308.
Hunk #3 FAILED at 1355.
```

PR #659, opened 2021-07-22
gsmlib: set C++ standard to 11

neheb:
throw() was removed in C++17, which GCC11 defaults to.

Maintainer: @micmac1 
Compile tested: ath79


micmac1:
Thank you!

PR #662, opened 2021-07-23
kamailio: replace patch with sed script

micmac1:
On the build bots the patch patches clashes with Build/Prepare. This
commit removes it in favor of a simple sed script. The resulting change
is identical.

Maintainer: @jslachta 
Compile tested: master mips64
Run tested: N/A (compile issue)

Description:
Hi Jiri,

As @neheb mentioned there's a problem with the recently introduced patch. It clashes with Build/Prepare on the build bots ("at home" the build works fine). This PR does away with the patch and replaces it with a sed script (same like we already do for the "-funroll-loops".

Kind regards,
Seb


neheb:
This issue I saw with an arch Linux container. I don't notice it with Fedora or Debian. I'll do some resting.

neheb (approved):

jslachta:
Hi Sebastian and Rosen,

thank you for proposed changes. I have not spotted this at my build machine on Debian. 

I acknowledge your changes.

Kind regards,
Jiri

jslachta (approved):

Issue #631, opened 2021-03-29
asterisk-chan-lantiq: when DTMF chan_lantiq.c:817 ast_lantiq_read: entering... no code here... and Channel TAPI/1 left 'simple_bridge' basic-bridge


jospezial:
Maintainer: @jslachta 
@kochstefan @arnysch can you help?
Environment: MIPS lantiq, VGV7510KW22 (o2 Box 6431) OpenWRT master head, asterisk-18.3.0, pjsip, chan-lantiq
Description:
After #597 was fixed, calls are terminated when sending a DTMF number.

```
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Got  RTP packet from    62.52.28.202:21512 (type 08, seq 014854, ts 077120, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Sent RTP packet to      62.52.28.202:21512 (type 08, seq 032290, ts 078080, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Got  RTP packet from    62.52.28.202:21512 (type 08, seq 014855, ts 077280, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Sent RTP packet to      62.52.28.202:21512 (type 08, seq 032291, ts 078240, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Got  RTP packet from    62.52.28.202:21512 (type 08, seq 014856, ts 077440, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: Sent RTP packet to      62.52.28.202:21512 (type 08, seq 032292, ts 078400, len 000160)
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: [Mar 29 13:25:47] DEBUG[3543]: chan_lantiq.c:1517 lantiq_dev_event_digit: on port 0 detected digit "5"
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: [Mar 29 13:25:47] DEBUG[3543]: chan_lantiq.c:1505 lantiq_send_digit: Port 0 transmitting digit "5"
Mon Mar 29 15:25:47 2021 daemon.info asterisk[2964]: [Mar 29 13:25:47] DTMF[6620][C-00000004]: channel.c:3866 __ast_read: DTMF end '5' received on TAPI/1, duration 0 ms
[...]

Issue #597, opened 2020-12-24
asterisk 18.1.0 crashes when a number is pressed (DTMF) when remote end is ringing or has picked up


jospezial:
Maintainer: @jslachta
Environment:
MIPS lantiq, VGV7510KW22 (o2 Box 6431) built from openwrt git master,Gigaset DECT on analog phoneports,chan_lantiq, pjsip,
Linux version 5.4.83 (joerg@jlgentoo) (gcc version 10.2.0 (OpenWrt GCC 10.2.0 r15241-3ab695368a)) #0 SMP Fri Dec 18 22:05:50 2020

Description:
I was having a call where I was prompted to input a number on the phone. After that the phone was quiet.
Reproducable also after reboot.

```Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Sent RTP packet to      62.52.130.135:41992 (type 08, seq 013240, ts 121920, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Got  RTP packet from    62.52.130.135:41992 (type 08, seq 019457, ts 3025006985, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Sent RTP packet to      62.52.130.135:41992 (type 08, seq 013241, ts 122080, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Got  RTP packet from    62.52.130.135:41992 (type 08, seq 019458, ts 3025007145, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Sent RTP packet to      62.52.130.135:41992 (type 08, seq 013242, ts 122240, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Got  RTP packet from    62.52.130.135:41992 (type 08, seq 019459, ts 3025007305, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Sent RTP packet to      62.52.130.135:41992 (type 08, seq 013243, ts 122400, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Got  RTP packet from    62.52.130.135:41992 (type 08, seq 019460, ts 3025007465, len 000160)
Thu Dec 24 12:22:12 2020 daemon.info asterisk[2648]: Sent RTP packet to      62.52.130.135:41992 (type 08, seq 013244, ts 122560, len 000160)
[...]

PR #663, opened 2021-07-31
asterisk: bump to 18.5.1

micmac1:
Maintainer: @jslachta 
Compile tested: master + 19.07 ath79
Run tested: 19.07 ath79

Description: Hi Jiri,

Minor bump + dep update + pjproject patch sync

Kind regards,
Seb

### DIFF
--- a/libs/pjproject/Makefile
+++ b/libs/pjproject/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pjproject
 PKG_VERSION:=2.10
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 # download "vX.Y.tar.gz" as "pjproject-vX.Y.tar.gz"
 PKG_SOURCE_URL_FILE:=$(PKG_VERSION).tar.gz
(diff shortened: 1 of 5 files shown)


dhewg:
Thanks, seems to work just fine for me!

micmac1:
Hi there @dhewg

If I'm not mistaken you're using lantiq. Do you use the phone ports and experience call releases with chan-lantiq when you press digits/DTMF, like reported in issue #631? I'm curious.

Kind regards,
Seb

dhewg:
Hi @micmac1,

nope, while indeed on lantiq arch, I'm not using chan-lantiq. I ran into the related #597 tough

micmac1:
But not anymore, right?

dhewg:
Correct, that was fixed with installing/depending on the timerfd asterisk module

micmac1:
Thanks!

jslachta:
Ack, compile and run tested.

Thank you, Sebastian!

jospezial:
@micmac1 
Please add PATCH 2/2 for
https://github.com/openwrt/telephony/blob/master/libs/pjproject/patches/0110-tls-parent-listener-destroyed.patch
See
https://github.com/dhewg/telephony/commit/501e539157e084f4cb09a1d9ee5bfae67c47d059
"pjproject: add missing patch from AST-2021-009
There are two related patches and only one was added."

Or a version bump to https://github.com/pjsip/pjproject/releases/tag/2.11.1 which contains this fix?

If I take a look at
https://github.com/pjsip/pjproject/security/advisories
there are some recent security vulnerabilities which are fixed in
https://github.com/pjsip/pjproject/commits/master

micmac1:
> Please add PATCH 2/2 for
> https://github.com/openwrt/telephony/blob/master/libs/pjproject/patches/0110-tls-parent-listener-destroyed.patch

Thanks for bringing this up. Added to both master and 21.02. 19.07
already had it.

PR #671, opened 2021-08-16
freeswitch: set stop level

micmac1:
This sets STOP to 10 to allow for a
graceful shutdown. Users may additionally need to set "term_timeout" in
/etc/config/freeswitch to a sensible value.

Maintainer: me
Compile tested: N/A
Run tested: 21.02 ath79

Description:
Update init script a bit after seeing some commits from stintel in "packages" repo. Makes sense to give FS a chance to clean up during shutdown.


micmac1:
Removed START level change. Don't want to cause any surprises to anybody.

PR #673, opened 2021-08-20
asterisk: update AST_EXT_TOOL_CHECK for OpenWrt

### LINKED ISSUE #672: asterisk picking up host libmariadb
Maintainer: @jslachta 
Environment: arc700

https://gist.github.com/neheb/f030945a8b63dbf6dbf941591103249c

micmac1:
Issue #672 shows that the current macro has some pitfalls, at least when
running within OpenWrt's buildroot.

The macro is used to detect these tools:

  1. xml2-config
  2. mysql_config
  3. neon-config
  4. net-snmp-config
  5. sdl-config (we're not using SDL)

The macro also adds configure arguments that allow to specify a path
like so:

  --with-mysqlclient=PATH

Macro problems:

  1. If no PATH is specified, it adds "/bin" to the search path.
     Obviously this is a bad thing when cross-compiling.
  2. If a PATH is specified, it adds it to the start of every include
     and library path it encounters. These paths are then broken.

This commit makes problem 1 go away by making the addition of
"${$1_DIR}/bin" to the PATH dependent on there being a PATH handed to
the macro. An alternative would be to remove the PATH addition
altogether, but then the macro wouldn't behave as expected.

It also removes the path filtering of the include and lib dirs to
address problem 2.

Closes #672

Maintainer: @jslachta 
Compile tested: 21.03 SDK
Run tested: Not tested, this is a build fix

Description:
Hi Jiri,

Hope you have good summer days!

Small macro amendment included. If you don't like the way I did it let me know, of course :)

No runtime change, so I didn't revbump.

Kind regards,
Seb

### DIFF
--- a/net/asterisk/patches/160-AST_EXT_TOOL_CHECK.patch
+++ b/net/asterisk/patches/160-AST_EXT_TOOL_CHECK.patch
@@ -1,0 +1,22 @@
+--- a/autoconf/ast_ext_tool_check.m4
++++ b/autoconf/ast_ext_tool_check.m4
+@@ -8,13 +8,16 @@ AC_DEFUN([AST_EXT_TOOL_CHECK],
+ 	AC_REQUIRE([AST_PROG_SED])dnl
+ 	if test "x${PBX_$1}" != "x1" -a "${USE_$1}" != "no"; then
+ 		PBX_$1=0
+-		AC_PATH_TOOL(CONFIG_$1, $2, No, [${$1_DIR}/bin:$PATH])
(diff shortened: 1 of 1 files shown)


micmac1:
Due to the way we use this macro the only changes for us are:

1. it will not pick up tools in /bin anymore
2. It fixes include and lib dirs for the res-snmp module

neheb:
Just tested. Fixed now.

jslachta (approved):

jslachta:
Thank you, Seb!

PR #675, opened 2021-08-21
gsmlib: disable NLS

micmac1:
It's not wanted, so add "--disable-nls" to override user selecting
CONFIG_BUILD_NLS.

configure.in and Makefile.am are slightly changed to make these
disappear:

ERROR: Use of AM_GNU_GETTEXT without [external] argument is no longer supported.

Makefile.am:21: warning: 'intl' should not be in SUBDIRS when AM_GNU_GETTEXT([external]) is used

Maintainer: me
Compile tested: ath79 master
Run tested: -

Description:

### DIFF
--- a/libs/gsmlib/Makefile
+++ b/libs/gsmlib/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gsmlib
 PKG_VERSION:=1.10-20140304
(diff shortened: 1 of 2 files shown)


neheb:
Works, ty.

micmac1:
Moved the autotools related part of the commit from a separate patch into existing patch 01-update-autotools.patch.

PR #693, opened 2021-09-30
asterisk: add opus read and write support

dhewg:
I've been using this for ~5 months now, works for me (tm).
In my case I store voicemail recording as .opus files and send those out via email.

There's currently no support to set the encoder properties using codecs.conf as I don't have a use case for it.
Noteworthy is that the encoding complexity is set to 4 (from the default of the max 10) on soft float devices, as they can't keep up encoding in realtime.

4 works for me on lantiq/xrx200, so a rather old and slow device.

### DIFF
--- a/net/asterisk-opus/Makefile
+++ b/net/asterisk-opus/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=asterisk-opus
-PKG_RELEASE:=1
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE_URL:=https://github.com/traud/asterisk-opus.git
-PKG_SOURCE_DATE:=2017-10-09
-PKG_SOURCE_VERSION:=83e1b458c77e0e287adeca494eeb79edb077b0ff
-PKG_MIRROR_HASH:=c71b859db7518cdafff1650e629c5901b290fe68f8af54ef1afd57bc9f15b122
+PKG_SOURCE_DATE:=2021-11-01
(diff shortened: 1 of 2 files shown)


traud:
Before you create a Pull Request with my upstream project, could you double-check whether something can be done about that CPU utilization aka floating issue? Opus Codec has a non-floating API. I have not checked if that
(a) can be used with Asterisk directly or
(b) Asterisk itself has to be compiled differently for such a platform.

traud:
Not sure I got that. I try again what I meant: Currently, you send the value `4` to `ope_encoder_ctl(…)` in the wrapper library libopusenc ([GitHub](https://github.com/xiph/libopusenc) and [Xiph.org](https://gitlab.xiph.org/xiph/libopusenc)). That calls the control function within the Opus Codec library itself. Now, libopusenc is using the floating point API of the Opus Codec library. However, the Opus Codec library has a fixed-point, software floating-point, or whatever you call it API as well. What happens, if you use that on one of your OpenWrt devices?

dhewg:
Updated, the patches are now upstream and not required here anymore ;)

micmac1:
Do you mind to make the following changes?

- instead of adding include path in TARGET_CFLAGS, add them to CPATH in
  Makefile patch, to prevent the build system adding include path
/usr/include/opus
- move Package/asterisk-codec-opus to
  Package/asterisk-opus/Default and then call it from
  Package/asterisk-codec-opus and Package/asterisk-format-ogg-opus and
  only change TITLE/DEPENDS to avoid duplication
- add a short description to asterisk-format-ogg-opus

Kind regards,
Seb

dhewg:
Sounds all good, done

micmac1:
Thank you!

Issue #685, opened 2021-09-16
 asterisk-chan-dongle: open('/var/lock/LCK..ttyUSB*') failed: Permission denied


532910:
Maintainer: @jslachta
Environment: cortexa53 friendlyarm_nanopi-neo2 21.02.0

Description:
1. Starting with 21.02.0 asterisk runs as `asterisk` user.
2. chowning `/dev/ttyUSB*` to `asterisk:dialout` manually gives another error:

```
ERROR[4594]: chan_dongle.c:140 lock_create: open('/var/lock/LCK..ttyUSB2') failed: Permission denied
WARNING[4594]: chan_dongle.c:225 opentty: unable to open /dev/ttyUSB2: Resource busy
```

micmac1:
I doubt /var/lock even exists. You might have to create it.

OpenWrt doesn't have a mechanism to set device permissions (think "udev" like) as far as I'm aware. So in my opinion this is not a problem of the asterisk package. What people do to get by is update the init script and set the permissions (or create directories like /var/lock/xyz) before the application is launched.

micmac1:
Actually you may be able to use OpenWrt hotplug for that [link](https://openwrt.org/docs/guide-user/base-system/hotplug).

532910:
I tried, but it doesn't provide device name (like /dev/ttyUSB0)

micmac1:
Provides other stuff that looks usable though, according to the documentation I linked. Like VENDOR and whatnot.

532910:
There is no VENDOR described on [hotplug](https://openwrt.org/docs/guide-user/base-system/hotplug) page.
I tried [troubleshooting](https://openwrt.org/docs/guide-user/base-system/hotplug#usagetroubleshooting) script and found no way to set /dev/ttyUSB0 permission.

micmac1:
It's in the PRODUCT var

rojer:
`/var/lock` does exist, actually, and is populated with other stuff:
```
root@SP-BR1:~# ls -l /var/lock
-rw-r--r--    1 root     root             0 Oct  1 19:59 procd_asterisk.lock
-rw-r--r--    1 root     root             0 Oct  1 19:55 procd_cron.lock
-rw-r--r--    1 root     root             0 Oct  1 19:59 procd_dnsmasq.lock
-rw-r--r--    1 root     root             0 Oct  1 19:55 procd_dropbear.lock
[...]

Issue #520, opened 2020-04-12
Asterisk 16.6.1 problems running Asterisk as user "asterisk" , chan_lantiq.c:2116 load_module: lantiq TAPI device open function failed


jospezial:
Maintainer: @jslachta 
Environment: MIPS lantiq, VGV7510KW22 (o2 Box 6431), OpenWrt SNAPSHOT, r12903-c30220d458 built from openwrt git master

Works with root/root in /etc/config/asterisk .
Which device is used? Should the all vmmc devices have dialout as group?

```
# If you have problems running Asterisk as user "asterisk" we'd
        # like to hear from you. Please raise an issue at:
        # https://github.com/openwrt/telephony/issues
        option user 'asterisk'
#       option user 'root'
        option group 'asterisk'
#       option group 'root'
```
```
cat /etc/group 
root:x:0:
daemon:x:1:
adm:x:4:
mail:x:8:
dialout:x:20:asterisk
audio:x:29:
www-data:x:33:
ftp:x:55:
users:x:100:
network:x:101:
nogroup:x:65534:
asterisk:x:385:asterisk
dnsmasq:x:453:dnsmasq
lp:x:7:p910nd
```

```Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:   == res_pjsip.so => (Basic SIP resource)
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:  Loading res_pjsip_authenticator_digest.so.
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:   == res_pjsip_authenticator_digest.so => (PJSIP authentication resource)
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:  Loading res_pjsip_endpoint_identifier_ip.so.
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:   == res_pjsip_endpoint_identifier_ip.so => (PJSIP IP endpoint identifier)
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:  Loading res_pjsip_endpoint_identifier_user.so.
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:   == res_pjsip_endpoint_identifier_user.so => (PJSIP username endpoint identifier)
Fri Apr 10 15:26:01 2020 daemon.info asterisk[2257]:  Loading res_pjsip_registrar.so.
[...]

PR #697, opened 2021-10-22
asterisk/asterisk-chan-dongle: address permission issues

micmac1:
Maintainer: @jslachta 
Compile tested: master
Run tested: the asterisk init script change I tested. Dongle changes untested by me, but I think some people had success with them.

Description:
Hello @532910 @rojer 

This adds the change suggested by @dhewg in #685, to use AST_RUN_DIR. This should sort out lock file creation.

This also removes setting the group from asterisk init script. So this should make the dongle accessible for the user, if the user is in the dialout group, for instance.

Regarding adding the user asterisk automatically to the group dialout. I decided against it. There is a function for this, "group_add_user", but no package uses it. Its sole user is "add_group_and_user". So adding the user to a group needs to be done manually, but that shouldn't be a big hassle.

I did add a postinst note to the package, that nudges users into the right direction (hopefully).

Does this work for you? I'd also like to backport this to 21.02. Let me know.

Thanks!
Seb

### DIFF
--- a/net/asterisk-chan-dongle/Makefile
+++ b/net/asterisk-chan-dongle/Makefile
@@ -13,7 +13,7 @@ PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/wdoekes/asterisk-chan-dongle.git
 PKG_SOURCE_VERSION:=3d2a97f26ba84473a0926a9fe9a32e451a56bc36
 PKG_SOURCE_DATE=2021-09-29
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_MIRROR_HASH:=b5baf737d39d522fa9b7055dea72c2d3241c272b20354fdf2a3bd09cce274844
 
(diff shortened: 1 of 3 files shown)


net/asterisk/files/asterisk.init
```diff
@@ -51,7 +51,6 @@ start_service() {
   procd_open_instance
   procd_set_param command $COMMAND
   procd_append_param command \
-    -G "$NAME" \
     -U "$NAME" \
```
jospezial, line 54: In my asterisk init script I also removed the line
    -U "$NAME" \
but have set in /etc/asterisk/asterisk.conf:
```
runuser = asterisk              ; The user to run as.
rungroup = dialout              ; The group to run as.
```
and in /etc/init.d/vmmc:
```
chown :dialout /dev/vmmc$i
```
Works for me since over a year together with git master.

See also:
https://github.com/openwrt/telephony/issues/520
and
https://github.com/openwrt/openwrt/pull/3124

PR #700, opened 2021-11-03
[19.07] bump pjproject, asterisk & chan-lantiq

micmac1:
Maintainer: @jslachta 
Compile tested: lantiq 19.07 SDK
Run tested: not yet

Description:
19.07 is going to be there for a while longer. I think we should update pjproject and asterisk to get the security fixes in. Just adding the patches would also be fine, but I'm not able to do that (not a programming specialist :)).

Anyway, for pjproject I used mostly cherry-picks, except for the 0111 patch.

The timerfd module is included into main package to prevent DTMF issues. I bumped chan-lantiq as well to get the DTMF fix there, too.

I can't downgrade my router this week. But next week should be fine. I'll give it a quick runtest then.

Hello @jospezial

Don't know if you're running 19.07 still somewhere to try chan-lantiq? If not no issue.

Kind regards,
Seb


jospezial:
> Don't know if you're running 19.07 still somewhere to try chan-lantiq? If not no issue.

No reason to for now. I'm on git master.

micmac1:
:-) the chance was pretty slim. Thanks anyway.

micmac1:
Squashed pjproject commits, updated wording up asterisk bump commit. Meantime downgraded my router to 19.07.8. So runtesting next.

micmac1:
Run-tested this on 19.07.7 ath79. Works fine, also DTMF. When I remove res-timing-timerfd it segfaults, so moving this module to core package also makes sense for 16.22.0. And considering this DTMF issue I also think that updating chan-lantiq makes sense.

Will merge this now.

PR #702, opened 2021-11-04
asterisk: add missing modules + dependency

micmac1:
Maintainer: @jslachta 
Compile tested: master ath79
Run tested: N/A

Description:
Add some new modules + a missing dependency that were missed. Will wait with this and revdump later, not urgent. Just adding now to not forget.


micmac1:
Just realized asterisk is now using $(AUTORELEASE), so no need to revbump.

PR #712, opened 2021-11-28
libsrtp: bump to 2.4.2

micmac1:
Maintainer: @jslachta 
Compile tested: master sdk from yesterday
Run tested: master sdk from yesterday, ath79

Description:
Bump of srtp lib. Runtested with asterisk and pjsip. Mobile client showed encryption was active, also signaling was proper in asterisk. Calls were fine, no audio issues.


jslachta:
Thank you, Sebastian.

PR #715, opened 2021-12-06
coturn: new package

micmac1:
This is meant as a replacement for restund.

By default, sqlite3 support is enabled. The DBs to use can be freely
changed in menuselect, though. sqlite3 is enabled by default as it is
the upstream default DB.

The package installs turnserver, which includes an init script. Options
can be set via /etc/config/turnserver and /etc/turnserver.conf.

turnserver is started as root and later drops privileges to user/group
turnserver/turnserver.

DB schemas etc. get installed to /usr/share/coturn. The default DB path
is set to /etc/turnserver, to avoid volatile /var on OpenWrt.
/etc/turnserver is created and permissions are automatically set to only
allow access for user turnserver (done via fakeroot).

The utilities coturn provides are installed as well.

Maintainer: me / @jslachta 
Compile tested: master/21.02
Run tested: master/21.02

Description: I used turnutils_uclient on Debian to test turnserver on OpenWrt. Seems to be OK.

Jiri can you do the last rites on restund, please? Move it to abandoned?

Kind regards,
Seb


micmac1:
Fixed the description of COTURN_ENABLE_REDIS flag.

PR #725, opened 2021-12-25
coturn: add libintl dependency when enabled full language support

BKPepe:
Maintainers: @jslachta and @micmac1 
Issue reproduced on these two routers:
- Turris MOX, mvebu, (cortex-a53), OpenWrt master
- Turris Omnia, mvebu (cortex-a9), OpenWrt master
and on the later one, I was able to reproduce this issue locally and fix it as well. I didn't had the build environment for the first one, yet. Thus proposing this fix.

If there is enabled CONFIG_BUILD_NLS, coturn can not be compiled.
It fails with the following error:

Package coturn is missing dependencies for the following libraries:
libintl.so.8

### DIFF
--- a/net/coturn/Makefile
+++ b/net/coturn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=coturn
 PKG_VERSION:=4.5.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/coturn/coturn/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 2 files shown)


micmac1:
> Issue reproduced on these two routers:
> - Turris MOX, mvebu, (cortex-a53), OpenWrt master
> - Turris Omnia, mvebu (cortex-a9), OpenWrt master
> and on the later one, I was able to reproduce this issue locally and
> fix it as well. I didn&#39;t had the build environment for the first
> one, yet. Thus proposing this fix.

Hi Josef,

Thanks for PR. I think adding the dep may not be best option here.
configure checks if it can link in intl, but nothing in coturn actually
uses it (nothing includes libintl.h, there's no gettext files either).

I'd suggest to just patch out the intl check from configure instead.

Thanks,
Seb

BKPepe:
Hi Seb!

Oh, that's too bad that I am not using coturn anyhow. I just noticed this build failure. I agree that my approach might not be ideal. It's good to know that you checked it that nothing uses intl in coturn. 👍 

I will look at what you said, but I can't probably perform a reliable run test. It would be good if this is fixed upstream as well.

micmac1:
I'll raise a PR upstream, no problem. Will link it here.

BKPepe:
Patch backported from the PR. Compile tested. Thank you. :+1:

micmac1:
Thank you Josef!

jslachta:
Thank you!

PR #734, opened 2022-02-01
coturn: improve reproducibility and fix build on macos

svlobanov:
This PR has two commits:
coturn: improve reproducibility (Ubuntu build host fix)
coturn: fix build on macos

Maintainer: @jslachta , Sebastian Kemper <sebastian_ml@gmx.net>
Compile tested: (armvirt/64, OpenWRT version b21bc3479d46e6a4c3cc6bf7c245d4b0ddccb7db)
Run tested: not really tested, but generated binaries are exactly the same as official binaries from https://downloads.openwrt.org/snapshots/packages/ . Now generated binaries are the same on Debian, Ubuntu and MacOS build hosts

CI status for first commit (reproducibility) is: https://github.com/svlobanov/telephony/runs/5024603191
CI status for both commits will be in this PR and by link: https://github.com/svlobanov/telephony/runs/5024839078

### DIFF
--- a/net/coturn/Makefile
+++ b/net/coturn/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=coturn
 PKG_VERSION:=4.5.2
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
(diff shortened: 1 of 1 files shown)


micmac1:
Thanks!

PR #735, opened 2022-02-01
restund: remove package

micmac1:
Removing as upstream dropped the project. Users can use coturn instead.

Maintainer: @jslachta 
Compile tested: NA
Run tested: NA

Description: drop package


jslachta:
Thank you, Seb.

PR #740, opened 2022-02-06
kamailio: fix build on macos

svlobanov:
kamailio macos build fails due to kamailio Makefile checks OS
(`uname -s`) and OSREL (`uname -r`). If build host is not Linux,
then these checks will provide different results, but OpenWrt is
always Linux so target OS should be always Linux.

This patch explicitly specifies OS=linux and OSREL=$(LINUX_UNAME_VERSION)
to avoid using build host values for target build.

Signed-off-by: Sergey V. Lobanov <sergey@lobanov.in>

Maintainer: @jslachta 
Compile tested: (armvirt/64, OpenWRT 4d904524effc9eb0cc5094574c55d3a520803223)

Description: see above

### DIFF
--- a/net/kamailio/Makefile
+++ b/net/kamailio/Makefile
@@ -235,6 +235,7 @@ PKG_CONFIG_DEPENDS:= \
 
 include $(INCLUDE_DIR)/nls.mk
 include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/kernel.mk
 include $(TOPDIR)/feeds/packages/lang/python/python3-version.mk
 
 # Build reproducibly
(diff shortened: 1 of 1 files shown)


micmac1:
Hello Sergey,

When including kernel.mk first the PKG_BUILD_DIR is different to what you would expect for a normal package. But we only want the LINUX_UNAME_VERSION, PKG_BUILD_DIR should remain as is.

So please include kernel.mk after package.mk:

```
include $(INCLUDE_DIR)/nls.mk
include $(INCLUDE_DIR)/package.mk
include $(INCLUDE_DIR)/kernel.mk
include $(TOPDIR)/feeds/packages/lang/python/python3-version.mk
```

Then this is fine.

Thanks!
Seb

micmac1:
Thank you!

PR #761, opened 2022-04-18
[21.02] bump asterisk to 18.11.2

micmac1:
Maintainer: @jslachta 
Compile tested: 21.02 SDK ath79
Run tested: 21.02.2 ath79

Description:
Hi all,

Applying AST-2022-00{1,2,3} patches to Asterisk 18.5.1 that we have in 21.02 branch is not possible (at least not for stir_shaken res modules) due to refactoring and other upstream changes. I guess changes could be made to the patches to make it work, but I wouldn't be able to do that myself.

So instead I updated asterisk to latest 18 LTS version, cherry-picking from master. I squashed the respective asterisk and pjproject commits so there's just two commits added to 21.02 branch in total.

Kind regards,
Seb

P.S.: will look at 19.07 branch as well


jslachta:
Thank you, Sebastian!

PR #771, opened 2022-06-07
rptengine: update to 10.4.1.4

PolynomialDivision:


net/rtpengine/patches/05-use-spandsp3.patch
```diff
@@ -9,7 +9,7 @@
  CFLAGS+=	-DWITH_TRANSCODING
  CFLAGS+=        $(shell mysql_config --cflags)
  else
-@@ -68,7 +68,7 @@ LDLIBS+=	$(shell pkg-config --libs libav
```
BKPepe, line 12: You are doing more changes than you are saying. You should add to the commit description than you refreshed patch.

PolynomialDivision: updatet

PR #773, opened 2022-06-12
freeswitch: workaround for failure with gcc 12

micmac1:
An issue report already exists upstream [1].

[1] https://github.com/signalwire/freeswitch/issues/1670

Maintainer: me
Compile tested: master ath79
Run tested: N/A

Description:
workaround for compile fail with gcc 12


micmac1:
On another note, I compiled all other telephony packages with gcc 12 and there were no failures, which is nice.

PR #775, opened 2022-07-16
asterisk-chan-dongle: adapt BUILD_NLS handling

micmac1:
openwrt recently stopped using libiconv-stub in the BUILD_NLS=n case.
Instead the libc's iconv has to be used.

This commit changes the iconv include path passed to
asterisk-chan-dongle for the BUILD_NLS=n case to
$(TOOLCHAIN_DIR)/include. That's where musl and glibc store iconv.h.

Also, the patch 200-fix-iconv-detection.patch gets replaced with a sed
script that only runs in the BUILD_NLS=y case.

And finally, $(AUTORELEASE) is added.

Maintainer: @jslachta 
Compile tested: ath79 master SDK
Run tested: N/A

Description: fix build errors in asterisk-chan-dongle due to recent removal of libiconv-stub from openwrt


jslachta (approved):

jslachta:
Thank you!

PR #779, opened 2022-10-07
Add esl-lua binding and build ESL.so [21.02]

pdepaulis:
Adds **freeswitch-misc-lua-esl** package to freeSWITCH 1.10.7-1 (OWRT 21.02).

Description:
A _freeswitch-misc-lua-esl_ package will be available in menuconfig after updating the feeds.
Select freeswitch-misc-lua-esl in the menuconfig to build it; the packages adds `/usr/lib/lua/ESL.so` which is the ESL Lua binding library.


pdepaulis:
This pull request is for `openwrt-21.02 3f098fb`

PR #788, opened 2022-11-04
pcapsipdump: fix linking against libobsd

guijan:
Maintainer: nobody
Compile tested: x64 glibc toolchain built from OpenWRT commit hash eb456aedfe24a076a4f53743ad3c090ae447329e with my PR at https://github.com/openwrt/openwrt/pull/9714 applied

Description:
```
This just makes it so libbsd is picked up through pkg-config. It adds
support for libobsd without breaking libbsd.
```
This is probably my first PR to OpenWRT that isn't fixing a bug while getting what I want done so I'm pinging @neheb as an interested party.

Signed-off-by: Guilherme Janczak <guilherme.janczak@yandex.com>

### DIFF
--- a/net/pcapsipdump/patches/01-use-libc-strlcpy-if-available-v2.patch
+++ b/net/pcapsipdump/patches/01-use-libc-strlcpy-if-available-v2.patch
@@ -1,49 +1,51 @@
 --- a/Makefile
 +++ b/Makefile
-@@ -2,8 +2,12 @@ LIBS ?= -lpcap -lstdc++
+@@ -2,11 +2,14 @@ LIBS ?= -lpcap -lstdc++
  RELEASEFLAGS ?= -O3 -Wall
  #CXXFLAGS ?= --std=c++0x
  
+-# auto-detect if bsd/strings.h is available
+-ifeq ($(shell $(CXX) $(CXXFLAGS) $(LDFLAGS) $(DEFS) -E -o /dev/null \
+-    	make-checks/libbsd.cpp 2>/dev/null; echo $$?),0)
+-	BSDSTR_DEFS := -DUSE_BSD_STRING_H
+-	BSDSTR_LIBS := -lbsd
 +# check if libc provides BSD's strlcpy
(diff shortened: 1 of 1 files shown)


neheb (approved):

neheb:
On an unrelated note, this package is weird. PKG_MIRROR_HASH is missing, mostly because SVN does not produce reproducible checkouts. It’s also the only SVN package. Probably should be fixed. Ping @micmac1

micmac1:
Thanks for the patch, looks good to me. I guess we can wait a bit till either libbsd is fixed or for libobsd to get added. I suppose with libobsd we wouldn't have the "cannot find /usr/lib/libbsd.so.0.11.7" issue, correct?

Hi @neheb, upstream is still on SF's SVN afaik, so we're a bit stuck here. I'm not even sure pcapsipdump is actually used by anybody running OpenWrt, also because there are alternatives like sipgrep/sngrep that see more upstream activity these days.

Kind regards,
Seb

micmac1:
Probably worth considering if maybe to do last rites on this package :)

neheb:
I second that suggestion.

guijan:
It shouldn't break libbsd, so don't commit it, I must have done something wrong...
And I'm indifferent to removing the package, I don't use this program.

neheb:
The issue is https://github.com/openwrt/openwrt/pull/11008

guijan:
The PR is fine as is, then.

micmac1:
> I second that suggestion.

OTOH guijan already did the work, so we can just wait for the libbsd stuff to get fixed in OpenWrt and keep pcapsipdump around. I'm not aware of any issues with it.

neheb:
Besides the PKG_MIRROR_HASH thing I mentioned, none.

micmac1:
> Besides the PKG_MIRROR_HASH thing I mentioned, none.

That item doesn't bother me much :D

neheb:
@micmac1 any chance to get this merged?

micmac1:
Hi Rosen, howdy! Will check tonight or tomorrow!

PR #792, opened 2022-12-02
[22.03] asterisk: bump to 18.15.1

micmac1:
Maintainer: @jslachta 
Compile tested: 22.03 sdk
Run tested: 22.03 ath79 made some test calls

Description:
Includes security fixes for both pj and asterisk. I tried just adding the patches, but that didn't work, some patch refers to some function that isn't included in 18.11.2. So I updated to 18.15.1 instead.

3 commits cherry-picked from master. Only the asterisk 18.15.1 commit is really new.


jslachta (approved):

jslachta:
Thank you!

PR #793, opened 2022-12-03
[21.02] asterisk: bump to 18.15.1

micmac1:
Maintainer: @jslachta 
Compile tested: 21.02 SDK
Run tested: downgraded my ath79 to 21.02 temporarily for giving this a quick run test. Test calls were OK.

Description:
This is the bump for the 21.02 branch to address a bunch of security concerns, same as for 22.03 branch, but without dropping the uclibc patches, as we still need them in the 21.02 branch.

As mentioned in the other pull request, I tried just adding the patches, but that didn't work, some patch refers to some function that isn't included in 18.11.2. So I updated to 18.15.1 instead.

Kind regards,
Seb


jslachta (approved):

jslachta:
Thank you!

Issue #796, opened 2023-01-09
rtpengine doesn't start


r-vit:
Maintainer: Sebastian Kemper <sebastian_ml@gmx.net>
Environment: arch: ARMv8 Processor rev 4, model: Xiaomi Redmi Router AX6S, OpenWrt version: 22.03.2 r19803-9a599fee93

Description:
rtpengine doesn't start with an error "Error relocating /usr/bin/rtpengine: lws_get_ssl: symbol not found"

package version is 9.5.3.2-2

micmac1:
Remove libwebsockets-mbedtls, install libwebsockets-full and report back, please.

r-vit:
it does work with this library

micmac1:
OK, will update the dependency.

micmac1:
Thanks for the report @r-vit 

Closing as done.

PR #798, opened 2023-01-09
[22.03] rtpengine: update libwebsockets dependency

micmac1:
The rtpengine daemon requires lws_get_ssl(). This function is not available in libwebsockets-mbedtls. This commit updates the dependency to libwebsockets-openssl.

Maintainer: me & @jslachta 
Compile tested: 22.03 ath79 sdk
Run tested: 22.03 with libwebsockets-full, but that should be the same as libwebsockets-openssl as far as lws_get_ssl() is concerned.

Description:
Update dependency as per #796

### DIFF
--- a/net/rtpengine/Makefile
+++ b/net/rtpengine/Makefile
@@ -48,7 +48,7 @@ ENGINE_DEPENDS := \
 	+libopenssl \
 	+libpcap \
 	+libpcre \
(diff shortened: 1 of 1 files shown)


jslachta (approved):

jslachta:
👍 , waiting for the build.

PR #803, opened 2023-02-10
freeswitch,coturn: Prepare for OpenSSL 3.0 upgrade

cotequeiroz:
Maintainer: @jslachta @micmac1 
Compile tested: arm, aarch64
Run tested: none

Description:
These are packages that need adjustments or version bumps to work with OpenSSL 3.0. See https://github.com/openwrt/openwrt/pull/11963 for its status.

Coturn needs a large patch, which I took from Gentoo, and appears to be backports from upstream.  I did not try to get too creative with the versions, but OpenSSL is currently supported in master, and the different "docker" releases, which I did not dig deep into, so I did not go farther than 4.6.1.


micmac1:
Thank you Eneas!

PR #819, opened 2023-07-09
[22.03] pjproject: bump to 2.13.1

micmac1:
Maintainer: @jslachta 
Compile tested: 22.03 SDK
Run tested: 23.05-rc2

Description:
Bump pjproject to include 3 security related fixes. Asterisk commit is to remove the libsrtp dep from res-pjproject. It's no longer required.


micmac1:
I was about to do the same for 21.02, but looking at the build bots it seems they haven't built any packages for 21.02 in a while. And then I found this [post](https://lists.openwrt.org/pipermail/openwrt-announce/2023-May/000041.html) from Hauke officially announcing 21.02 is EOL on 1st of May.

PR #820, opened 2023-07-18
treewide: replace AUTORELEASE with real PKG_RELEASE

kuanyili:
Maintainer: @micmac1

Description:

Automatically compute and substitute current values for all $(AUTORELEASE) instances as this feature is deprecated and shouldn't be used.

All affected packages are untouched since 23.05 branching. It should be straightforward applying this patch also to that branch.


micmac1:
Thanks Kuan-Yi,

libsrtp should be at 2, chan-dongle and asterisk-opus likewise, rtpproxy
at 4. Rest seems fine. Nice catch with the sngrep double PKG_RELEASE!

Can you correct this?

Kind regards,
Seb

kuanyili:
@micmac1

Ah! It turns out I forgot to switch the repo from shallow clone to full clone
before running the command, should be fine new.

Thanks for taking the time to review this patch!

micmac1:
Thanks a bunch!

PR #829, opened 2023-08-31
Add CPE ID for asterisk and pjproject

Rondom:
Maintainer: @micmac1 
Compile tested: x86 / 64 / master
Run tested: no run tests done, given the nature of the change it should be alright

Description:
Add CPE IDs for Asterisk and pjproject


jslachta:
Hello Andreas (@Rondom),

Ack from me. Thank you for your contribution.

Kind regards,
Jiri

PR #830, opened 2023-09-11
siproxd: Configuration option registration_file setting was not applied

schuettecarsten:
The configuration setting registration_file was not correctly applied to the generated siproxd-general.conf file. Whenever this option was set in siproxd config file, it was completely missing in the generated conf file. This fixes this.

### DIFF
--- a/net/siproxd/files/siproxd.init
+++ b/net/siproxd/files/siproxd.init
@@ -19,6 +19,20 @@ append CONF_SKIP "interface_inbound interface_outbound chrootjail"
 append CONF_SKIP "daemonize user plugindir registration_file pid_file"
 
 
+# Check if a UCI option is set and append it to config
+
+apply_conf() {
+	local var="$1"
+	local sec="$2"
+	local opt="$3"
+	local val
+
+	config_get "$var" "$sec" "$opt"
+	eval "val=\"\${$var}\""
+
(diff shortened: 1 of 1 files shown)


guidosarducci:
> The configuration setting registration_file was not correctly applied to the generated siproxd-general.conf file. Whenever this option was set in siproxd config file, it was completely missing in the generated conf file. This fixes this.

The setting is not missing; it's there and managed internally. This is by design, to ensure proper setup of procd, jails, and UCI section handling, and also explicitly stated in the the code:
```
# Some options need special handling or conflict with procd/jail setup.
append CONF_SKIP "interface_inbound interface_outbound chrootjail"
append CONF_SKIP "daemonize user plugindir registration_file pid_file"
```

There's no bug I can see that's fixed by this PR, but it might help if you articulated your use case or reasons more clearly.

schuettecarsten:
As written in the original description: The settings `registration_file` in the file `/etc/config/siproxd` was not applied into the generated `siproxd.conf` file. So the file is always written to the default location. I want to put the file on a mounted USB volume to make sure that it is reboot-safe.

But because siproxd is really unefficient and unstable with kernel 6.1, I found a different solution and removed siproxd from my setup. So if you do not see a a problem here, feel free to close this pull request.

PR #834, opened 2023-10-10
yate: Update yate script to use an nftables set

robho:
Maintainer: @jslachta
Compile tested: -
Run tested: Netgear R6220, 23.05.0-rc3

By using an nftables set in this script it's easier to install and use this script now that OpenWrt uses nftables by default.


micmac1:
Thanks Robert!

PR #786, opened 2022-10-23
dahdi-linux: bump + fixes for linux>=5.18 and 6.1 build

john-tho:
Maintainer: 
Compile tested: aarch64 kernel 6.1rc1 WIP
Run tested: no

wait for upstream changes PR review: https://github.com/asterisk/dahdi-linux/pull/14

PR #836, opened 2023-10-13
dahdi-linux: bump to latest git head

micmac1:
This compiles with newer kernels now.

Maintainer: N/A
Compile tested: sdk-ipq807x (kernel 6) sdk-ath79 (kernel 5)
Run tested: N/A

Description:
Hi @ john-tho 

I tried upstream master and it compiles now. Are you OK with this?

Kind regards,
Seb


Ansuel:
mhhh can't we bump to a stable release and backport the needed patches? the error with std is a missing configure option so it should be just a patch to backport (if it's not already included in 3.2.0)

micmac1:
It's quite a lot of patches, even if you update to 3.2.0, see PR #786. If john-tho updates his pull request I'd be fine with this approach, though, obviously.

In my opinion this is a legacy package with limited upstream support, so keeping our efforts as minimal as possible seems like a measured approach.

voltagex:
I hit this compile error without explicitly asking for it to be built - how can I disable it so that my build can succeed for the moment?

micmac1:
Try disabling the VoIP kernel modules. Or maybe uninstall the package temporarily (./scripts/feeds uninstall dahdi-linux) and rerun menuconfig.

PR #839, opened 2023-11-02
rtpengine: bump to 11.5.1.12 release and set PCRE2

Ansuel:
Bump rtpengine to release 11.5.1.12.

New 11.x release require libopus as a new library and also moved to PCRE2 library.

Refresh patches and also introduce a new one to disable docs as they changed doc generation tool and now would require a new prereq and docs are not useful and would waste space on an embedded scenario.

Description:


micmac1:
Thanks Christian

PR #682, opened 2021-09-08
asterisk-chan-quectel:Add asterisk support for Quectel EC25 modem

mperic02:
Quectel EC25 modem support for asterisk.

Signed-off-by: Matej Peric <peric.matej1@gmail.com>

PR #862, opened 2024-04-17
Openwrt 22.03

huanghitoy:

Description:


micmac1:
Hi there,

Please see [1], we already had this discussion. I guess same applies still: Please send any fixes to chan-dongle upstream and then we can import them here.

I'll probably close this pull request next week if nobody objects.

Kind regards,
Seb

[1] https://github.com/openwrt/telephony/pull/682

PR #882, opened 2024-08-18
sofia-sip: removed "--disable-stun"

olivluca:
Maintainer: @micmac1 
Run tested: (mips, Astoria Networks ARV7518PW,  OpenWrt 23.05.4 r24012-d8dd03c46f , sip registration and made a test sip call)

Description:

I'm trying to update my [svd package](https://github.com/olivluca/danube-voip) for OpenWrt 23.05.4. To compile it I need stun enabled. ~~I'm working on a test repository [here](https://github.com/olivluca/danube-voip-test) where~~  In my repository I adopted this version of the Makefile.

### DIFF
--- a/libs/sofia-sip/Makefile
+++ b/libs/sofia-sip/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=sofia-sip
 
 PKG_VERSION:=1.13.17
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


jslachta (approved):

micmac1:
I don't see a point for this. The software that is using this is not in this repo. The user can keep the change locally. Not needed for openwrt telephony.

Kind regards,
Seb

olivluca:
But why `--disable-stun `in the first place?

micmac1:
Am 21. September 2024 19:25:46 UTC schrieb Luca Olivetti ***@***.***>:
>But why `--disable-stun `in the first place?
>

Freeswitch doesn't need it for STUN.

olivluca:
But would it work with it? It only adds 540 bytes to the compiled library.

micmac1:
Ah, so it's only a tiny increase. Thanks for pointing that out. Then I don't mind.

Kind regards,
Seb

PR #890, opened 2024-11-18
gsmlib: Make version apk compatible

hnyman:
Adjust version to be compatible with apk's semantic rules.
Currently gsmlib is failing in buildbot, and also then apparently also breaking freeswitch.
https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/telephony/gsmlib/compile.txt

```
ERROR: info field 'version' has invalid value: package version is invalid
ERROR: failed to create package: /builder/shared-workdir/build/sdk/bin/packages/aarch64_cortex-a53/telephony/gsmlib-1.10-20140304-r5.apk: package version is invalid
make[3]: *** [Makefile:122: /builder/shared-workdir/build/sdk/bin/packages/aarch64_cortex-a53/telephony/gsmlib-1.10-20140304-r5.apk] Error 99
```

Needed change should be easy, as the actual code specification is by git hash, and version is just information.

Maintainer: nobody
cc @Ansuel  @jslachta @micmac1 
Compile tested: hopefully CI


hnyman:
CI apparently misses gettext-full...

compile-tested ok for mediatek/mt7622 RT3200

```
perus@ub2410:/OpenWrt/e8450$ ls bin/packages/aarch64_cortex-a53/telephony/
gsmlib-1.10.20140304-r5.apk  index.json  packages.adb  tmp
```

hnyman:
anybody still here with merge rights?

jslachta:
Yup, here. Going to merge this today.

jslachta:
Hi Hannu (@hnyman),

I believe changing the PKG_VERSION variable will result in a different package checksum. I’m not entirely sure, but in this case, I would suggest bumping the PKG_RELEASE.

Otherwise, it looks good to go.

Kind regards,
Jiri

hnyman:
> I believe changing the PKG_VERSION variable will result in a different package checksum. I’m not entirely sure, but in this case, I would suggest bumping the PKG_RELEASE.

Note that "major version item" PKG_VERSION changes, so the "minor version item" PKG_RELEASE is actually irrelevant. I left that unchanged at 5, but it also could be argued to be resetted back to 1 (as PKG_VERSION did change).

But if you want, I can change that to 6.

jslachta:
Thank you!

PR #895, opened 2025-05-22
asterisk: update to version 20.14.0

pprindeville:
Maintainer: @jslachta 
Compile tested: x86_64, generic, HEAD
Run tested: same, installed on production IP-PBX

Description:

### DIFF
--- a/net/asterisk/Makefile
+++ b/net/asterisk/Makefile
@@ -8,13 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=asterisk
-PKG_VERSION:=20.8.1
+PKG_VERSION:=20.14.0
 PKG_RELEASE:=1
 PKG_CPE_ID:=cpe:/a:digium:asterisk
 
 PKG_SOURCE:=asterisk-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.asterisk.org/pub/telephony/asterisk/releases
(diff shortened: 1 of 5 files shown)


pprindeville:
@neheb @Ansuel Any idea why the Intel builds are failing?

robimarko:
dahdi-linux is failling to compile, so the whole build fails

pprindeville:
> You probably just need to update dahdi-linux as well (which is by no means x86/intel only btw).

Okay, thanks. Working on it.

It seems that there's a requirement for `kmod-echo` that isn't being satisfied.

Also, where's the documentation for building, refreshing, etc. an out-of-tree kernel module?

pprindeville:
> Also, where's the documentation for building, refreshing, etc. an out-of-tree kernel module?

That is, `make -j1 package/kmod-dahdi-linux/{clean,download,refresh} V=sc` doesn't seem to do it.

pprindeville:
@robimarko Following your advice about the version bump.  By the way, how do I build out-of-tree kernel modules with extra CFLAGS like `-Wno-old-style-declaration`?  And per @dangowrt, this isn't strictly x86_64 so how do I enable it for the other applicable platforms (guessing ARM)?

pprindeville:
> You probably just need to update dahdi-linux as well (which is by no means x86/intel only btw).

Got a separate PR but the CI/CD system failing to set up the build environment for `x86_64` for some reason.

See PR #904.

pprindeville:
Thanks!

PR #920, opened 2025-09-21
asterisk: add ssl and crypto build options

james-knippes:
Maintainer: @micmac1 
Compile tested: lantiq/xway (openwrt-24.10) , lantiq/xrx200 (openwrt-24.10, main)
Run tested: lantiq/xway (openwrt-24.10) , lantiq/xrx200 (openwrt-24.10)

Description: These options allow compiling asterisk with or without crypto and ssl support. This is mainly useful for image size optimization for small devices.

### DIFF
--- a/net/asterisk/Makefile
+++ b/net/asterisk/Makefile
@@ -396,6 +396,19 @@ define Package/$(PKG_NAME)/config
 		  Warning: this feature is known to cause problems with some modules.
 		  Disable it if you experience problems like segmentation faults.
 
+	config ASTERISK_SSL_SUPPORT
+		bool "Build asterisk with SSL support"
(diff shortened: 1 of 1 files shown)


james-knippes:
Would this qualify for being backported to 24.10 ? I'm not sure if we could bend [the rules](https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md#release-branches) and qualify this as 'fix' as this essentially fixes building asterisk for devices with small flash sizes.

Ansuel:
@james-knippes i wonder if it can be compiled with only crypto... is it even possible? Doesn't that also require libopenssl (for the libcrypto?)

PR #927, opened 2025-12-03
asterisk: bring up to current - bump from 20.14.1 to 22.2.7->23.1.0

systemcrash:
This patch-set fixes asterisk compile and brings it up to current. See individual commit messages for details. 

@jslachta 

This will likely fix other asterisk components which depend on asterisk and fail to build.

### DIFF
--- a/libs/pjproject/Makefile
+++ b/libs/pjproject/Makefile
@@ -10,7 +10,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pjproject
-PKG_VERSION:=2.14.1
+PKG_VERSION:=2.15.1
 PKG_RELEASE:=1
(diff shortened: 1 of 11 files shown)


systemcrash:
Asterisk and all modules compile and load cleanly (and only errors/warnings resulting from lack of config are emitted).

systemcrash:
Note to self: asterisk 23 / master named the sample config phoneprov_users.conf instead of .sample so it doesn't install during make. The user can manually provide it if necessary until it's fixed in master and the next version bump here.

See https://github.com/asterisk/asterisk/pull/1627

Ansuel:
@systemcrash can you fix the module for one of the asterisk ?

systemcrash:
@Ansuel did not understand.

PR #930, opened 2025-12-26
rtpengine: update to 13.5.1.3

graysky2:
Maintainer: no one
Compile tested: x86/64
Run tested: not run tested
Description:
This release is needed to build against linux 6.18.

Removed upstreamed patches:
101-use-spandsp3.patch
104-daemon-poller-fix-compilation-error-for-pthread-tid.patch

Manually rebased:
100-prevent-systemd-detection.patch

Declare a dependency on BTREE_SUPPORT for rtpengine userspace variants and ensure the kernel module enables CONFIG_BTREE when built.

This PR depends on https://github.com/openwrt/openwrt/pull/21416


net/rtpengine/patches/101-disable-config_btree_warn.patch
```diff
@@ -0,0 +1,10 @@
+--- a/kernel-module/xt_RTPENGINE.c
++++ b/kernel-module/xt_RTPENGINE.c
+@@ -38,7 +38,6 @@
+ #include <linux/btree.h>
+ #define KERNEL_PLAYER
+ #else
+-#warning "Kernel without CONFIG_BTREE - kernel media player unavailable"
+ #endif
+ 
+ #include "xt_RTPENGINE.h"
```
BKPepe: It is good that you do have mention why this patch is needed in the pull request, but it is missing in the patch itself.

BKPepe: To be honest, this patch sounds to me like a hack and I dont like it so much.

graysky2: This was just my idea to solve the build error. Happy to go another direction.

BKPepe: > the module enforces -Werror, causing the build to fail on OpenWrt kernels where CONFIG_BTREE is disabled. 

Do you have any log to see what is going on? I dont think that this should be merged as it is. 
It is a question, why the issue is raised suddenly in 13.5.1.3. 🤔

graysky2:
```
make[5]: Entering directory '/scratch/union/build_dir/target-x86_64_glibc/rtpengine-no-transcode/rtpengine-mr13.5.1.3/kernel-module'
  CC [M]  xt_RTPENGINE.o
xt_RTPENGINE.c:41:2: error: #warning "Kernel without CONFIG_BTREE - kernel media player unavailable" [-Werror=cpp]
   41 | #warning "Kernel without CONFIG_BTREE - kernel media player unavailable"
      |  ^~~~~~~
cc1: all warnings being treated as errors
```

systemcrash: This looks fine, but should probably be fixed (in a better way) upstream. We can carry this for the time being.

jslachta: Instead of removing the warning, would it make sense to extend [target/Config.in](https://github.com/openwrt/openwrt/blob/main/target/Config.in) with BTREE support and make **rtpengine** depend on `BTREE_SUPPORT`?

graysky2: I think that make sense. I updated the PR. I do not think it is quite right. When I try building this I am seeing:
```
make[3]: Entering directory '/scratch/union/build_dir/target-x86_64_glibc/rtpengine-no-transcode/rtpengine-mr13.5.1.3/recording-daemon'
../utils/gen-common-flags >../config.mk.new
Error: Required package 'json-glib-1.0' not found
../lib/deps.Makefile:15: ../config.mk: No such file or directory
make[3]: *** [../lib/deps.Makefile:6: ../config.mk] Error 1
make[3]: Leaving directory '/scratch/union/build_dir/target-x86_64_glibc/rtpengine-no-transcode/rtpengine-mr13.5.1.3/recording-daemon'
```

Further the kernel config is not getting the `CONFIG_BTREE=y` option.
```
% grep BTREE /scratch/union/build_dir/target-x86_64_glibc/linux-x86_64/linux-6.12.63/.config
CONFIG_IPV6_SUBTREES=y
# CONFIG_RBTREE_TEST is not set
```

PR #934, opened 2026-02-20
asterisk: bump to 23.2.2

systemcrash:
dropped patch 001 - integrated at source

all patches refreshed


systemcrash:
ping @jslachta

dhewg:
I did the same just yesterday ;) It may be worth mentioning the fixed CVEs?

I also have this, and I wouldn't mind if you just include it in your PR:
https://github.com/openwrt/telephony/commit/a3545592bafd253f83b203423add1994d191f4b3

systemcrash:
Shall I throw everything else in there?

dhewg:
> Shall I throw everything else in there?

Oh, you just took 'em all, hah!
For this bump I'd really just add the single commit I pointed to, as it fixes actual package bugs.

The other stuff is still useful IMHO, but:

I had a PR about these a few years ago, but they were rejected:
> asterisk: clean up init.d script
> asterisk: add a reregister extra command
> asterisk: add an 'interface' option to trigger reregister events

These two patch asterisk, but as I won't sign upstream's agreement they won't be merged there and hence we need to maintain them if accepted:
> asterix: fix build with CONFIG_OPENSSL_NO_DEPRECATED
> asterisk: VM_MESSAGEFILE

Which is why they don't have a proper commit message and are not even signed off...

systemcrash:
We’re basically carrying the other patches in perpetua …

systemcrash:
ping... @Ansuel ?

jslachta:
Hi @systemcrash ,

the last commit - asterix: fix build with CONFIG_OPENSSL_NO_DEPRECATED - fails on formalities - missing SoB line.

Otherwise, it looks good to me. Thank you so much! I am sorry for the delay.

systemcrash:
OK - fixed those.

jslachta:
It fails on formalities. I am merging this anyway. Thank you so much for your hard work.

PR #946, opened 2026-09-13
treewide: rework libatomic dependencies with ATOMIC64_DEPENDS

dangowrt:
Maintainer: rtpproxy Jiri Slachta, libks and signalwire-client-c have no maintainer listed
Compile tested: ramips/mt7621 and x86/64 snapshot SDKs (r36222) with openwrt/openwrt#25112 applied; every package passes the packaging dependency check on both
Run tested: none, build and packaging tests only

Description:

Follow-up to openwrt/openwrt#25112, which added `HAS_ATOMIC64` and `ATOMIC64_DEPENDS:=+!HAS_ATOMIC64:libatomic`: the compiler calls libatomic for 64-bit atomics only on mips, mipsel, powerpc and ARM below ARMv6K. This reworks the three packages in this feed that depend on libatomic according to what their built binaries import (`readelf -DWs` on the packaged files):

- rtpproxy imports `__atomic_fetch_add_8`, `__atomic_load_8` and `__atomic_store_8` on mips: `+libatomic` becomes `$(ATOMIC64_DEPENDS)`.
- libks and signalwire-client-c link libatomic on every architecture without importing a symbol from it: the dependency is dropped and both link with `--as-needed`, so no `NEEDED libatomic.so.1` remains.


dangowrt:
CI note: the failing build jobs on the 32-bit targets (mips, mipsel, powerpc, ARMv5) are expected until the snapshot SDKs include openwrt/openwrt@e23f575e4d2f3732faa5ec6e8486bc41ac3d7b25. The current SDKs (r36237 and older) predate that commit, so `ATOMIC64_DEPENDS` expands to nothing there, the packages lose their libatomic dependency and the packaging check rejects the `NEEDED libatomic.so.1`, exactly the failure mode this change is meant to make impossible once the core side is in. The jobs will be re-run when the next snapshots are out.

dangowrt:
Correction for this feed: the 32-bit jobs here passed even with the pre-merge SDKs because the packaging check accepts libraries provided by any transitive dependency and libsrtp2 pulls in libopenssl, which already depends on libatomic on those targets. The actual failures are unrelated to the change: mips_24kc, arm_cortex-a15_neon-vfpv4 and x86_64 died with `wget: exited with error 8` while fetching the test container, and aarch64_generic and i386_pentium-mmx fail the generic version test because `rtpproxy -V` prints the release tag from rtpp_version.h (`rel.20160514172346`) and makeann and extractaudio print no version at all, so no executable matches `2025.11.04~4909e7b7`. That test has never passed for rtpproxy; it needs a `test-version.sh` in the package directory.

